PR #19, opened 2020-03-31
Using real context to shutdown metrics HTTP server. 

HvonP:
Using real context to shutdown metrics HTTP server. A nil context is not safe towards the http package and will on occasion cause panic

**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**
Using real context to shutdown metrics HTTP server. A nil context is not safe towards the http package and will on occasion cause panic

### DIFF
--- a/metrics/transport.go
+++ b/metrics/transport.go
@@ -119,7 +119,7 @@ func PullMetrics(ctx context.Context, metricsPort int, actions []string) error {
 	go func() {
 		<-ctx.Done()
 		//nolint:staticcheck
-		if err := srv.Shutdown(nil); err != nil {
+		if err := srv.Shutdown(context.Background()); err != nil {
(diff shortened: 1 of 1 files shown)


metrics/transport.go
```diff
@@ -119,7 +119,7 @@ func PullMetrics(ctx context.Context, metricsPort int, actions []string) error {
 	go func() {
 		<-ctx.Done()
 		//nolint:staticcheck
-		if err := srv.Shutdown(nil); err != nil {
+		if err := srv.Shutdown(context.Background()); err != nil {
```
SeBBBe: According to docs the Shutdown can potentially hang indefinitely. We should probably have something like:
context.WithTimeout(context.Background(), 5*time.Second)

PR #24, opened 2020-04-07
Enhancement/569 change reuse user randomizer

DnlLrssn:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

* Make session id count also for reuse user iterations
* Determinism not broken due to a failed iteration when using reuse user flag
* Bugfix for determinsm and thinktime actions
* Randomizer can now be overriden by other by users or future schedulers

### DIFF
--- a/generatedocs/data/actions/thinktime/examples.md
+++ b/generatedocs/data/actions/thinktime/examples.md
@@ -2,7 +2,7 @@
 
 #### ThinkTime uniform
 
-This would simulate a think time between 10 and 15 seconds.
+This simulates a think time of 10 to 15 seconds.
 
 ```json
 {
@@ -18,7 +18,7 @@ This would simulate a think time between 10 and 15 seconds.
 
 #### ThinkTime constant
 
-This would simulate a think time of 5 seconds.
+This simulates a think time of 5 seconds.
 
 ```json
 {
--- a/generatedocs/data/actions/thinktime/examples.md
+++ b/generatedocs/data/actions/thinktime/examples.md
@@ -2,6 +2,8 @@
 
 #### ThinkTime uniform
 
+This would simulate a think time between 10 and 15 seconds.
+
 ```json
 {
      "label": "TimerDelay",
@@ -15,6 +17,8 @@
 ```
 
 #### ThinkTime constant
+
+This would simulate a think time of 5 seconds.
 
 ```json
 {
--- a/helpers/distribution.go
+++ b/helpers/distribution.go
@@ -2,12 +2,10 @@ package helpers
 
 import (
 	"fmt"
-	"strconv"
-	"math/rand"
-
 	"github.com/pkg/errors"
 	"github.com/qlik-oss/gopherciser/enummap"
-	"github.com/qlik-oss/gopherciser/randomizer"
+	"strconv"
+	"time"
 )
 
 type (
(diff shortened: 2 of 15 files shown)


helpers/distribution.go
```diff
@@ -89,21 +89,22 @@ func (settings DistributionSettings) Validate() error {
...
-		return settings.Delay, nil
+		return time.Duration(settings.Delay * float64(time.Second)), nil
 	case UniformDistribution:
-		min := settings.Mean - settings.Deviation
-		max := settings.Mean + settings.Deviation
-		delay := (rand.Float64() * (max-min)) + min
-		return delay, nil
+		min := time.Duration(float64(time.Second) * (settings.Mean - settings.Deviation))
+		max := time.Duration(float64(time.Second) * (settings.Mean + settings.Deviation))
+		return rnd.RandDuration(min, max)
```
SeBBBe, line 100: I was thinking about this, whether deviation should correspond to (max - min) or whether the value should be +/- deviation. But I'm not sure deviation is a well defined property of this distribution.

Apparently the standard deviation is  (max - min) / sqrt(12)...

At any rate, the documentation for thinktime is very sparse on details so maybe we should have another look at that.

scheduler/simple.go
```diff
@@ -147,22 +146,6 @@ func (sched SimpleScheduler) iteratorNewUsers(ctx context.Context, timeout time.
 	return errors.WithStack(helpers.FlattenMultiError(mErr))
 }
 
-func (sched SimpleScheduler) iteratorReuseUsers(ctx context.Context, timeout time.Duration, log *logger.Log,
```
SeBBBe, line 150: Maybe iteratorNewUsers should change names to just iterator or iterateUsers since it now handled both cases with new users for every session as well as reusing users, depending on its input parameters.

helpers/distribution.go
DnlLrssn: Yes standard deviation is something different. Can't see how max-min could be interpreted as how much it deviates? Since the other value specified is mean, it's, to me, obvious to be deviation from the mean, but will take a look how it's written in documentation.

scheduler/simple.go
DnlLrssn: ah sure, good point

helpers/distribution.go
SeBBBe: Yeah maybe, like I said I'm not really sure what I'd expect it to be. Both sort of makes sense to me. Right now it's (max - min) = 2 * deviation which I guess is what you'd prefer then.

So the value can deviate by _deviation_ from the mean. But individual samples can deviate from each other by up to 2 * _deviation_

DnlLrssn: Made some small changes to docs, most importantly to think time example.

DnlLrssn:
@carlcervin remember, these two files are generated:
* docs/settingup.md
* generatedocs/generated/documentation.go

generatedocs/data/actions/thinktime/examples.md
```diff
@@ -2,6 +2,8 @@
 
 #### ThinkTime uniform
 
+This would simulate a think time between 10 and 15 seconds.
```
carlcervin, line 5: Maybe change to "This simulates a think time of 10 to 15 seconds."?

generatedocs/data/actions/thinktime/examples.md
```diff
@@ -16,6 +18,8 @@
 
 #### ThinkTime constant
 
+This would simulate a think time of 5 seconds.
```
carlcervin, line 21: Maybe change to "This simulates a think time of 10 to 15 seconds."?

generatedocs/data/actions/thinktime/examples.md
DnlLrssn: changed

generatedocs/data/actions/thinktime/examples.md
DnlLrssn: changed

PR #25, opened 2020-04-08
ElasticMoveApp action

SeBBBe:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Add new ElasticMoveApp action to move apps between spaces.


scenario/elasticmoveapp.go
```diff
@@ -0,0 +1,107 @@
...
+type (
+	// ElasticMoveAppSettings settings for moving an app between spaces
+	ElasticMoveAppSettings struct {
+		session.AppSelection
+		DestinationSpace
+	}
+
+	DestinationSpace struct {
+		DestinationSpaceId   string `json:"destinationspaceid" displayname:"New space ID" doc-key:"destinationspace.destinationspaceid"`
+		DestinationSpaceName string `json:"destinationspacename" displayname:"New space name" doc-key:"destinationspace.destinationspacename"`
```
DnlLrssn, line 23: Add comments

DnlLrssn, line 23: Add comments

scenario/elasticuploadapp.go
```diff
@@ -27,15 +27,17 @@ type (
...
 		CanAddToCollection
 	}
 
 	CanAddToCollection struct {
 		Title      session.SyncedTemplate `json:"title" displayname:"Title" doc-key:"elasticuploadapp.title"`
 		Stream     session.SyncedTemplate `json:"stream" displayname:"Stream name" doc-key:"elasticuploadapp.stream"`
 		StreamGUID string                 `json:"streamguid" displayname:"Stream ID" doc-key:"elasticuploadapp.streamguid"`
-		// Deprecated: This property will be removed.ElasticShareApp action should be used instead, keeping entry here for some time to make sure scripts get validation error.
+		// Deprecated: This property will be removed. ElasticShareApp action should be used instead, keeping entry here for some time to make sure scripts get validation error.
 		Groups []string `json:"groups" displayname:"Groups - deprecated" doc-key:"canaddtocollection.groups"`
```
DnlLrssn, line 41: should update this reference as share app will disappear. Maybe even time to remove the Groups check?

scenario/elasticuploadapp.go
```diff
@@ -199,8 +210,8 @@ func (settings ElasticUploadAppSettings) Execute(sessionState *session.State, ac
 		}
 	case Legacy:
 		parameters := ""
-		if settings.SpaceID != "" {
-			parameters = fmt.Sprintf("?spaceId=%v", settings.SpaceID)
+		if destSpace != nil {
+			parameters = fmt.Sprintf("&spaceId=%v", destSpace.ID)
```
DnlLrssn, line 214: should be `?` here not `&`

scenario/elasticmoveapp.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func (settings DestinationSpace) ResolveDestinationSpace(sessionState *session.State) (*elasticstructs.Space, error) {
+	var moveToSpace *elasticstructs.Space
+	var err error
+	if settings.DestinationSpaceId != "" {
+		moveToSpace, err = sessionState.ArtifactMap.GetSpaceByID(settings.DestinationSpaceId)
+	} else if settings.DestinationSpaceName != "" {
+		moveToSpace, err = sessionState.ArtifactMap.GetSpaceByName(settings.DestinationSpaceName)
```
DnlLrssn, line 101: This would require a preceeding elasticexplore action where the space has been looked up. We should probably use the function `searchForSpaceByName` from elastiexplore instead, this will look in artifactmap first, otherwise search for it.

scenario/elasticuploadapp.go
SeBBBe: I was thinking about that as well. Might be time.

scenario/elasticuploadapp.go
SeBBBe: Good catch

generatedocs/data/params.json
```diff
@@ -1,4 +1,10 @@
 {
+    "destinationspace.destinationspaceid": [
+        "Destination space, specified by ID."
```
carlcervin, line 3: Maybe change to "ID of the destination space."?

generatedocs/data/params.json
```diff
@@ -1,4 +1,10 @@
...
+    ],
+    "destinationspace.destinationspacename": [
+        "Destination space, specified by name."
```
carlcervin, line 6: Maybe change to "Name of the destination space."?

generatedocs/data/params.json
SeBBBe: Sure, I could do that. I think maybe the current wording more suggests that only one of them should be used, as they specify the same thing in two different ways.

SeBBBe: Your wording seems to suggest to me that both should be typed. Agree/disagree? It's up to you :)

carlcervin: You've got a point. 
Maybe we should go with "Specify destination space by ID." and "Specify destination space by name."?

SeBBBe: I like that one

PR #34, opened 2020-05-04
Fix unmarshal bug for poll interval in elasticreload

SeBBBe:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

Fix unmarshal bug for poll interval in elasticreload
###

### DIFF
--- a/helpers/timeduration_test.go
+++ b/helpers/timeduration_test.go
@@ -1,0 +1,19 @@
+package helpers
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestMarshalTimeDuration(t *testing.T) {
+	td := TimeDuration(2000000000)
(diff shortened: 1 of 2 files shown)


markuswall:
Works fine for me, I see the 10s setting propagate properly to the script in the log.

helpers/timeduration_test.go
```diff
@@ -0,0 +1,22 @@
+package helpers
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestMarshalTimeDuration(t *testing.T) {
```
DnlLrssn, line 8: Wouldn't hurt to make a marshal/unmarshal test for the entire action (and all other actions), we should probably add that as a task to be done.

PR #36, opened 2020-05-06
metrics: Add gopherciser_build_info metric

hairyhenderson:
It's conventional to have a `*_build_info` metric (a gauge always set to `1`), with labels to indicate the current-running version. See https://www.robustperception.io/exposing-the-software-version-to-prometheus for more information on this convention.

We had some confusion around which version was running (due to missing metrics), and this would've helped us troubleshoot faster.

Signed-off-by: Dave Henderson <David.Henderson@qlik.com>

### DIFF
--- a/metrics/custom.go
+++ b/metrics/custom.go
@@ -4,11 +4,16 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	promNS = "gopherciser"
+)
+
 // GopherActions action counter
 var GopherActions = prometheus.NewCounterVec(
 	prometheus.CounterOpts{
-		Name: "gopherciser_actions_total",
-		Help: "Number of gopherciser actions and their result. DOES NOT USE LABEL!",
+		Namespace: promNS,
+		Name:      "actions_total",
+		Help:      "Number of gopherciser actions and their result. DOES NOT USE LABEL!",
 	},
 	[]string{"result", "action"},
 )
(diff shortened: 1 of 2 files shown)


metrics/custom.go
```diff
@@ -4,49 +4,59 @@ import (
...
+const (
+	promNS = "gopherciser"
+)
+
 // GopherActions action counter
 var GopherActions = prometheus.NewCounterVec(
 	prometheus.CounterOpts{
-		Name: "gopherciser_actions_total",
-		Help: "Number of gopherciser actions and their result. DOES NOT USE LABEL!",
+		Namespace: promNS,
```
HvonP, line 14: Namespace will be hardcoded to "promNS" independent of where or how we run this.
How will it be used?

SeBBBe:
LGTM but I'm not that familiar with Prometheus.

metrics/custom.go
DnlLrssn:
```golang
promNS = "gopherciser"
```
so gopherciser, but yes don't know how namespace effects either

hairyhenderson: @HvonP it's already hardcoded to `gopherciser` effectively...

`Namespace` is used as a common prefix to the final name. In this particular case, since Gopherciser is a specific application, it doesn't make sense to ever use a different one. But it reduces duplication.

There's also a `Subsystem` key which can be used sometimes, but isn't really necessary in this particular case.

See the docs on those here: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#Opts

PR #44, opened 2020-05-13
Improved cookie handling

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Improved cookie handling

### DIFF
--- a/enigmahandlers/sensedialer.go
+++ b/enigmahandlers/sensedialer.go
@@ -53,6 +53,7 @@ func setupDialer(dialer *enigma.Dialer, timeout time.Duration) {
 			return nil, errors.WithStack(err)
 		}
 
+		wsHeader := httpHeader.Clone()
 		if dialer.Jar != nil {
 			// cookie needs to be set using http not ws scheme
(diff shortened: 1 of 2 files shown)


enigmahandlers/sensedialer.go
```diff
@@ -69,13 +70,13 @@ func setupDialer(dialer *enigma.Dialer, timeout time.Duration) {
 				}
 			}
 			if len(cookieStrings) > 0 {
-				httpHeader.Add("Cookie", strings.Join(cookieStrings, "; "))
+				wsHeader.Add("Cookie", strings.Join(cookieStrings, "; "))
```
SeBBBe, line 73: So we want to add the cookies only for the dialer but not keep them further?

DnlLrssn: They are usually in the cookiejar, which is supported by most other things, this websocket implementation however doesn't support cookiejar so need to add them to the rest of the headers manually. But yes, the cookie header should only be added here.

PR #46, opened 2020-05-25
Feature/appstructure consumption

SeBBBe:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Refactor app structure logic to avoid cyclic imports
* Add interface and implementations to consume app structure
* Port bug fixes from bugfix/590-AppStructure

### DIFF
--- a/appstructure/appstructure.go
+++ b/appstructure/appstructure.go
@@ -153,3 +153,31 @@ func (err AppStructureObjectNotFoundError) Error() string {
 func (err AppStructureNoScenarioActionsError) Error() string {
 	return "no applicable actions in scenario"
 }
+
+// GetSelectables get selectable objects from app structure
+func (structure *AppStructure) GetSelectables(rooObject string) ([]AppStructureObject, error) {
+	rootObj, ok := structure.Objects[rooObject]
+	if !ok {
+		return nil, AppStructureObjectNotFoundError(rooObject)
+	}
+
+	return structure.addSelectableChildren(rootObj), nil
+}
+
+func (structure *AppStructure) addSelectableChildren(obj AppStructureObject) []AppStructureObject {
+	selectables := make([]AppStructureObject, 0, 1)
+	if obj.Selectable {
+		selectables = append(selectables, obj)
+	}
+
+	for id := range obj.Children {
+		child, ok := structure.Objects[id]
+		if !ok {
+			continue
+		}
+
+		selectableChildren := structure.addSelectableChildren(child)
+		selectables = append(selectables, selectableChildren...)
+	}
+	return selectables
+}
--- a/appstructure/appstructure.go
+++ b/appstructure/appstructure.go
@@ -1,0 +1,155 @@
+package appstructure
+
+import (
+	"encoding/json"
+	"github.com/qlik-oss/gopherciser/enummap"
+)
+
+type (
+
+	// MetaDef meta information for Library objects such as dimension and measure
+	MetaDef struct {
+		// Title of library item
+		Title string `json:"title,omitempty"`
+		// Description of library item
+		Description string `json:"description,omitempty"`
+		// Tags of  of library item
+		Tags []string `json:"tags,omitempty"`
+	}
+
+	// AppObjectDef title and ID of a Sense object
+	AppObjectDef struct {
+		// Id of object
+		Id string `json:"id"`
+		// Type of Sense object
+		Type string `json:"type"`
+	}
+
+	AppStructureMeasureMeta struct {
+		// Meta information, only included for library items
+		Meta *MetaDef `json:"meta,omitempty"`
+		// LibraryId connects measure to separately defined measure
+		LibraryId string `json:"libraryId,omitempty"`
+		// Label of on measure
+		Label string `json:"label,omitempty"`
+		// Def the actual measure definition
+		Def string `json:"def,omitempty"`
+	}
+
+	AppStructureDimensionMeta struct {
+		// Meta information, only included for library items
+		Meta *MetaDef `json:"meta,omitempty"`
+		// LibraryId connects dimension to separately defined dimension
+		LibraryId string `json:"libraryId,omitempty"`
+		// LabelExpression optional parameter with label expression
+		LabelExpression string `json:"labelExpression,omitempty"`
+		// Defs definitions of dimension
+		Defs []string `json:"defs,omitempty"`
+		// Labels of dimension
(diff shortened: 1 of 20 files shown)


scenario/deletesheet.go
```diff
@@ -20,7 +21,7 @@ type (
 	DeleteSheetSettings struct {
 		DeletionMode SheetDeletionModeEnum `json:"mode" displayname:"Deletion mode" doc-key:"deletesheet.mode"`
 		Title        string                `json:"title" displayname:"Sheet title" doc-key:"deletesheet.title"`
-		ID           string                `json:"id" displayname:"Sheet ID" doc-key:"deletesheet.id"`
+		ID           string                `json:"id" displayname:"Sheet ID" doc-key:"deletesheet.id" appstructure:"sheet"`
```
DnlLrssn, line 24: same here, should probably be `appstructure:"object"`

scenario/openapp.go
```diff
@@ -304,3 +305,21 @@ func (connectWs connectWsSettings) Execute(sessionState *session.State, actionSt
...
+	for _, obj := range structure.Objects {
+		if obj.Type == "sheet" {
+			newObjs.Sheets = append(newObjs.Sheets, obj)
+		}
+	}
+	for _, v := range structure.Bookmarks {
+		newObjs.Bookmarks = append(newObjs.Bookmarks, v)
+	}
+	return &newObjs, nil, true
+}
```
DnlLrssn, line 325: openapp should not add any active objects other than bookmarks

appstructure/appstructure.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	// AppStructurePopulatedObjects is the type returned by an action when prompted for selectable objects
+	AppStructurePopulatedObjects struct {
+		// Parent id of the parent object
+		Parent string
+		// Objects first level app objects returned by the current action
+		Objects []AppStructureObject
+		// Sheets objects of sheet type
+		Sheets []AppStructureObject
```
DnlLrssn, line 116: There's no reason to separate sheets, it's just an object with a hierarchy, same as e.g. filterpane.

scenario/changesheet.go
```diff
@@ -21,7 +22,7 @@ import (
 type (
 	// ChangeSheetSettings settings for change sheet action
 	ChangeSheetSettings struct {
-		ID string `json:"id" displayname:"Sheet ID" doc-key:"changesheet.id"`
+		ID string `json:"id" displayname:"Sheet ID" doc-key:"changesheet.id" appstructure:"sheet"`
```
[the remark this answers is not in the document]
DnlLrssn: Ok, understand why you want this here now, to filter out only this type. In that case it's fine, but the filtered type shouldn't be hardcoded as it's own parameter in AppStructurePopulatedObjects, it needs to be more generic so any object type can be filtered, e.g. `pivot-table` or any custom object type. Filter applied on object dictionary using visualization. Question is, how do we separate between applying filter on active objects or all objects?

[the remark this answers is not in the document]
SeBBBe: It's kinda hard coded, but kinda not. From an API perspective, you're free to specify any object type, including pivot-table etc. In that case it will filter from the object map automatically.

There's no support for non-active objects.

scenario/openapp.go
SeBBBe: It should add its sheets to the list of currently available sheets?

scenario/applybookmark.go
```diff
@@ -79,3 +80,37 @@ func (settings ApplyBookmarkSettings) Execute(sessionState *session.State, actio
...
+func (settings ApplyBookmarkSettings) AffectsAppObjectsAction(structure appstructure.AppStructure) (*appstructure.AppStructurePopulatedObjects, []string, bool) {
+	id := settings.BookMarkSettings.ID
+	if id == "" { // No ID, specified, search by title
+		title := settings.BookMarkSettings.Title.String()
+		for _, v := range structure.Bookmarks {
+			if v.Title == title {
+				id = v.ID
+				break
+			}
+		}
```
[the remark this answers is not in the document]
SeBBBe: Yes, I'm aware of this. No way around it, I guess.

scenario/openapp.go
DnlLrssn: changesheet should filter on all instead of on active.

scenario/deletesheet.go
SeBBBe: What, why? Deletesheet should only display objects of "sheet" type.

scenario/openapp.go
SeBBBe: I'm not following the distinction. If there are multiple apps in a script, it should only show the sheets of the currently open app, i.e "active"?

scenario/changesheet.go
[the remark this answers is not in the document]
DnlLrssn: Since we know subscribeobject is upcoming within the coming days, there's no need to implement it in a way that we need to throw away almost directly.

scenario/deletesheet.go
DnlLrssn: wrote this before my comment on changesheet.

scenario/changesheet.go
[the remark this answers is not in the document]
SeBBBe: I still don't follow. Subscribeobject won't affect this implementation. Changesheet will still only change sheets, not other types of objects.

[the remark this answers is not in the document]
SeBBBe: The only thing that appstructure:"sheet" does is tell the GUI that the changesheetsettings.ID should have a dropdown with all the active sheets in it, which won't change? Even if I merge the sheet and object lists as suggested, it doesn't have any bearing on this API.

scenario/openapp.go
DnlLrssn: only the current app yes. But there's a list of "active objects" and a dictionary of "all objects". Instead of handling sheets separate, we need a way for an action to tell if the filter types, in the case of changesheet this would be `sheet` should be applied on the list of active objects (i.e. "added" by an action) or on the dictionary of all objects in the app.

scenario/changesheet.go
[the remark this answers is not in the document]
DnlLrssn: I still don't see why this is hardcoded, What if I implement my own type of container, Then make an activate container action. I would need the same kind of logic, doesn't make sense to hardcode it.

PR #47, opened 2020-05-28
Enhancement/593 object handling

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add `make verify` Makefile command (executes quickbuild, test and lint-min)
* Move uniqueInts to helper package
* Move getting object functions and setting up get data events from changesheet to session
* Add functionality to register special handling of object type
* Add auto-chart handling as special handling
* Get Select paths from new objecthandler


helpers/uniqueInts_test.go
```diff
@@ -0,0 +1,60 @@
...
+		t.Fatalf("uniqueInts array was nil")
+	}
+
+	validateIntArraySortable(t, "uniqueInts", a, []int{5, 22}, true)
+
+	if !m.HasValue(5) {
+		t.Error("uniqueInts collection doesn't contain expected value: 5")
+	}
+
+	if m.HasValue(10) {
```
SeBBBe, line 30: that seems.. specific?

session/objecthandling.go
```diff
@@ -0,0 +1,741 @@
...
+		getObject := func(ctx context.Context) error {
+			var err error
+			genObj, err = sense.CurrentApp.Doc.GetObject(ctx, name)
+			return err
+		}
+		if err := sessionState.SendRequest(actionState, getObject); err != nil {
+			return errors.Wrapf(err, "Failed go get object<%s>", name)
+		}
+
+		obj, err := sense.AddNewObject(genObj.Handle, enigmahandlers.ObjTypeSheetObject, name, genObj)
```
SeBBBe, line 90: I know it's old but maybe the name "ObjTypeSheetObject" isn't ideal

session/objecthandling.go
```diff
@@ -0,0 +1,741 @@
...
+		rawGeneratedProp, errDataPath := generatedPropPath.Lookup(rawAutoChartProperties)
+		if errDataPath != nil {
+			return errors.Wrapf(errDataPath, "Failed to get generated properties for autochart<%s>", autochartGen.GenericId)
+		}
+
+		// Create sessionObject of type
+		var genObj *enigma.GenericObject
+		createSessionObject := func(ctx context.Context) error {
+			var err error
+			genObj, err = uplink.CurrentApp.Doc.CreateSessionObjectRaw(ctx, rawGeneratedProp)
```
SeBBBe, line 177: Nice! This is much better than before

session/objecthandling.go
```diff
@@ -0,0 +1,741 @@
...
+		} else {
+			isStackedDims = 0
+		}
+
+		datapages, err := gob.GetHyperCubeReducedData(ctx, requestDef.Path, []*enigma.NxPage{
+			{
+				Left:   isStackedDims,
+				Top:    0,
+				Width:  len(hypercube.DimensionInfo) + len(hypercube.MeasureInfo) + isStackedDims,
+				Height: int(math.Min(2000.0, 10000.0/float64(len(hypercube.MeasureInfo)+1))),
```
SeBBBe, line 407: Some magic numbers here. Could do with an explanation/constants.

helpers/uniqueInts_test.go
DnlLrssn: old file that was moved, I think it's just a test to make sure it doesn't always return true.

DnlLrssn: i.e. value doesn't matter

session/objecthandling.go
DnlLrssn: Agree, didn't do this type of refactoring. Maybe `ObjTypeGenericObject` works better?

SeBBBe: Yeah I think that better describes it

PR #52, opened 2020-06-03
avoid Get data warning for auto-chart main object

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Avoid warnings for getting data on auto-chart main object.

### DIFF
--- a/session/autocharthandler.go
+++ b/session/autocharthandler.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"context"
+	"github.com/pkg/errors"
 	"github.com/qlik-oss/enigma-go"
 	"github.com/qlik-oss/gopherciser/action"
 	"github.com/qlik-oss/gopherciser/enigmahandlers"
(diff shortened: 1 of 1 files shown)


session/autocharthandler.go
```diff
@@ -23,9 +24,33 @@ func (handler *AutoChartHandler) Instance(id string) ObjectHandlerInstance {
...
+	instance.GetObjectLayout(sessionState, actionState, obj)
+	handleAutoChart(sessionState, actionState, genObj, obj)
+}
+
+// GetObjectLayout for auto-chart main object
+func (instance *AutoChartInstance) GetObjectLayout(sessionState *State, actionState *action.State, obj *enigmahandlers.Object) {
 	sessionState.QueueRequest(func(ctx context.Context) error {
-		return GetObjectLayout(sessionState, actionState, obj)
-	}, actionState, true, "")
+		enigmaObject, ok := obj.EnigmaObject.(*enigma.GenericObject)
```
SeBBBe, line 34: Assume this is a different object than `genObj *enigma.GenericObject` in the call to `SetObjectAndEvents`?

DnlLrssn: auto-chart main object will always be *enigma.GenericObject

DnlLrssn:
> Don't really understand what this change does

auto-chart main object should only sen a GetLayout. This was previously achieved by having a object definition with auto-chart set to getLayout. without this definition and using the generic GetObjectLayout we would get a warning such as  `Get Data for object type<auto-chart> not supported`, protocolwise the correct messages where sent though. But since main auto-chart object only ever send GetLayout, and the rest of the logic is on the session object, this is a stripped-dpwn version of GetObjectLayout

session/autocharthandler.go
DnlLrssn: ah, understand the question now, no should be the same, logic was copied from the generic version, I could probably send in genObj instead, yes

SeBBBe:
That makes sense. Thx for the explanation

session/autocharthandler.go
SeBBBe: Doesn't really matter. But that means this code and the error check are not strictly needed.

DnlLrssn: you're right, didn't check it through for redundant things.

PR #60, opened 2020-06-17
Update ODAG generation actions to work with June 2020

SeBBBe:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Update ODAG generation actions to work with June 2020

### DIFF
--- a/scenario/generateodag.go
+++ b/scenario/generateodag.go
@@ -246,11 +246,11 @@ func getSelectionStateFromBinding(binding elasticstructs.OdagLinkBinding, sessio
 	var dataPages []*enigma.NxDataPage
 	err = sessionState.SendRequest(actionState, func(ctx context.Context) error {
 		var err error
-		dataPages, err = obj.GetListObjectData(ctx)
+		dataPages, err = obj.GetHypercubeData(ctx)
 		return err
 	})
 	if err != nil {
-		actionState.AddErrors(errors.Wrapf(err, "failed to get listobject data for field <%s>", binding.SelectAppParamName))
+		actionState.AddErrors(errors.Wrapf(err, "failed to get hypercube data for field <%s>", binding.SelectAppParamName))
 	}
 	for _, dataPage := range dataPages {
 		for _, listObjectValue := range dataPage.Matrix {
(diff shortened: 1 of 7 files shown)


senseobjects/odagtoolbarnavpoint.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+)
+
+func (otn *OdagToolbarNavpoint) setLayout(layout *OdagToolbarNavpointLayout) {
+	otn.mutex.Lock()
+	defer otn.mutex.Unlock()
+	otn.layout = layout
+}
+
+// UpdateLayout get and set a new layout for sheetlist
```
DnlLrssn, line 33: not sheetlist

scenario/generateodag.go
```diff
@@ -372,12 +379,12 @@ func GetOdagSelectionBindings(host string, odagLinkId string, sessionState *sess
...
-// createFieldListboxAsync creates a listbox session object for specified field
-func createFieldListboxAsync(sessionState *session.State, actionState *action.State, doc *enigma.Doc, field string) (*senseobjects.ListBox, error) {
-	var obj *senseobjects.ListBox
+// createOdagToolbarNavpointAsync creates a listbox session object for specified field
+func createOdagToolbarNavpointAsync(sessionState *session.State, actionState *action.State, doc *enigma.Doc, field string) (*senseobjects.OdagToolbarNavpoint, error) {
+	var obj *senseobjects.OdagToolbarNavpoint
 	err := sessionState.SendRequest(actionState, func(ctx context.Context) error {
 		var err error
-		obj, err = senseobjects.CreateListBoxObject(ctx, doc, field)
+		obj, err = senseobjects.CreateOdagToolbarNavpoint(ctx, doc, field)
```
DnlLrssn, line 387: Is a new session object created for every ODAG request in client? E.g. considering this scenario
1. selects x,y
2. Create ODAG
3. select v,w
4. Create ODAG
Should 2 session objects be created or is the object in (2) updated through normal update mechanism and updated during (3)?

SeBBBe: They are always created (and instantly destroyed after the request has been sent). So it follows the client flow.

senseobjects/odagtoolbarnavpoint.go
SeBBBe: Will fix

scenario/generateodag.go
DnlLrssn: ah I see when looking in full code, didn't show in diff, makes sense then

PR #63, opened 2020-06-18
use updated go-tus fork for now

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Update go-tus dependency to fork using latest gotusd. Solves issues with `go mod tidy`. Should be updated to main project once merged, but has not happened yet and branch is since march 14.


DnlLrssn:
> 
> 
> Approved, but we should revert back to original repo once fixed

which is exactly what I wrote in the PR subject

PR #74, opened 2020-07-15
Log elastic reload duration consistently as milliseconds

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Log elastic reload duration consistently as milliseconds to make it easier to parse by analyzer.


DnlLrssn:
> 
> 
> LGTM, but what could yield an error with ReloadTime parsing?

string not formatted as time.Duration parsable, which it should be according to reload team (that or empty). So should never happen, but we all know about "should"...

PR #75, opened 2020-07-15
Apicompliance/616 Openhub

alfinhaji:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Several API calls have been updated from limit=30 to limit=24, as well several new additional API calls, as a part of QCS elasticopenhub action.

### DIFF
--- a/elasticstructs/datafilesresponse.go
+++ b/elasticstructs/datafilesresponse.go
@@ -1,0 +1,41 @@
+package elasticstructs
+
+type DataFilesResp struct {
+	Data  []DataFilesRespData `json:"data"`
+	Links DataFilesRespLinks  `json:"links"`
+}
+
+type DataFilesRespQCredentials struct {
+	Ciphertext string `json:"ciphertext"`
+}
+
+type DataFilesRespData struct {
+	Created              string                    `json:"created"`
+	DatasourceID         string                    `json:"datasourceID"`
+	ID                   string                    `json:"id"`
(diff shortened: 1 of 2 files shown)


alfinhaji:
I built both oss and trial versions with this change so far and both built and were able to run elasticopenhub successfully, and validated the API changes with traffic logs

DnlLrssn:
> 
> 
> @DnlLrssn I do need assistance on capturing the data connection ID in the response for the new dc-dataconnections API call. Need to parse response and get the "id" for "qName": "DataFiles" which is then used in the subsequent /api/v1/qix-datafiles?top=1000&connectionId=xxxx call (which I have commented out for now)

There seems to be 24 hits with `"qName": "DataFiles"` when I check my case, so qName is not the selector, needs to be something else. will try to find out.

DnlLrssn:
ok seems to be close to what it is actually. 
it's `"qName": "DataFiles"` without a `space` parameter

DnlLrssn:
Added the data struct as `elasticstructs/datafilesresponse.go`, wasn't sure if you wanted to try yourself to add the logic too use it.

DnlLrssn:
changes so far lgtm.

DnlLrssn:
also merged latest master into the branch to make build pass

alfinhaji:
Tested both oss and internal, with and without jwtlogin action, and with both tenant admin and test users, passed all this testing.

PR #77, opened 2020-07-16
ListBoxSelect Action

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Add ListBoxSelect action**

 Adds a listbox select action for performing selections specifically available for list objects.

Closes #557

### DIFF
--- a/scenario/listboxselect.go
+++ b/scenario/listboxselect.go
@@ -103,8 +103,6 @@ func (settings ListBoxSelectSettings) Execute(sessionState *session.State, actio
 		actionState.AddErrors(errors.Errorf("Expected generic object , got object type<%T>", t))
 		return
 	}
-
-	sessionState.Wait(actionState)
 }
 
 func (settings ListBoxSelectSettings) doSelect(sessionState *session.State, actionState *action.State, genericObj *enigma.GenericObject) {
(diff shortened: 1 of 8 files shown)


scenario/listboxselect.go
```diff
@@ -0,0 +1,178 @@
...
+		selectFunction = genericObj.SelectListObjectExcluded
+	case Possible:
+		selectFunction = genericObj.SelectListObjectPossible
+	default:
+		actionState.AddErrors(errors.Wrapf(err, "Unknown filter pane selection type ListBoxSelectType<%s>", settings.Type))
+		return
+	}
+
+	selectListObjects := func(ctx context.Context) error {
+		sessionState.LogEntry.LogDebugf("Select in object<%s> h<%d> type<%s>", genericObj.GenericId, genericObj.Handle, genericObj.GenericType)
```
DnlLrssn, line 140: type of select might be good info in this debug too.

scenario/listboxselect.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if settings.Wrap {
+		beginSelections := func(ctx context.Context) error {
+			return genericObj.BeginSelections(ctx, []string{selectPath})
+		}
+		sessionState.QueueRequest(beginSelections, actionState, false, fmt.Sprintf("Failed to select in %s", genericObj.GenericId))
+		sessionState.Wait(actionState)
+		if actionState.Errors() != nil {
+			return
+		}
```
DnlLrssn, line 163: I guess this is copy-pasted, but we have some changes which were implemented since then. 
* There's a actionState.Failed that can be checked instead of Errors()
*  Wait function will return a bool if it was failed or not, so no need to check actionState.Failed in this specific case
* There's now also sessionState.SendRequest which is Sync (we used to only have QueueRequest).

scenario/listboxselect.go
```diff
@@ -0,0 +1,178 @@
...
+
+		sessionState.Wait(actionState)
+		if actionState.Errors() != nil {
+			return
+		}
+
+		sessionState.QueueRequest(endSelections, actionState, true, fmt.Sprintf("Failed to end selection in %s", genericObj.GenericId))
+	}
+}
```
DnlLrssn, line 178: you should end the action with a `sessionState.Wait`, as it is now, if wrap=false (default) it will exit the action before the selectListObjects request is finished.

scenario/listboxselect.go
atluq: I'll just change everything to Sync then. Good point on just looking at the return value if I would keet async.

scenario/listboxselect.go
atluq: The caller was waiting, but changing to sync anyways.

DnlLrssn: can't see anything was waiting for sessionState.QueueRequest(selectListObjects, actionState, false, fmt.Sprintf("Failed to select in %s", genericObj.GenericId))

independently, it's not only the request itself we need to wait for, each request response _might_ contain a list of changed objects, requests to update data for these objects will be automatically triggered async int he background. So to wait for there we need to end with a Wait.

scenario/listboxselect.go
atluq: Fixed

scenario/listboxselect.go
atluq: Alright! There was a third wait last in Execute function at line 107. If this does not wait for sessionState.QueueRequest(selectListObjects, actionState, false, fmt.Sprintf("Failed to select in %s", genericObj.GenericId)) in the case of wrap==false I am misunderstanding something.

DnlLrssn: ah sorry github diff confusing me, for some reason thought this was end of Execution, should have checked it out

DnlLrssn: Don't see anything there.

DnlLrssn: so should still end the action with a Wait for everything that wasn't triggered by "us" directly.

atluq: Okay, that I did no know, so I removed it. Will put the ending Wait in place again.

Can we be sure that triggered events will not affect the state for the next request, such that Wait in between request are not needed?

DnlLrssn: we don't need to wait for them in this case no, the "reaction" request are fetching layout and data for the objects on the sheet. If we send multiple request effecting the state of an object the engine will send an "aborted" answer to the data request which was ongoing, then the mechanism we have for reaction will re-send it again. Thus getting updated state after the second request.

DnlLrssn: none of the background request will do changes to the actual state in engine, they will however trigger the data to be calculated for the current state.

atluq: Ahh, I'm learning ;)

PR #80, opened 2020-07-17
Move Enum interface

atluq:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Move Enum inteface to actionhandler.go
* Implement Enum interface for ListBoxSelectType

### DIFF
--- a/scenario/actionhandler.go
+++ b/scenario/actionhandler.go
@@ -15,6 +15,7 @@ import (
 	"github.com/qlik-oss/gopherciser/appstructure"
 	"github.com/qlik-oss/gopherciser/buildmetrics"
 	"github.com/qlik-oss/gopherciser/connection"
+	"github.com/qlik-oss/gopherciser/enummap"
 	"github.com/qlik-oss/gopherciser/helpers"
 	"github.com/qlik-oss/gopherciser/logger"
 	"github.com/qlik-oss/gopherciser/session"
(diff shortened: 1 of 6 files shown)


scenario/listboxselect.go
```diff
@@ -69,7 +72,7 @@ func (value ListBoxSelectType) MarshalJSON() ([]byte, error) {
 
 // String representation of ListBoxSelectType
 func (value ListBoxSelectType) String() string {
-	sType, err := listBoxSelectTypeEnumMap.String(int(value))
+	sType, err := value.GetEnumMap().String(int(value))
```
DnlLrssn, line 75: I would prefer if we didn't allocate a new object for each user each time this is executed since it never changes anyway.
you could keep the listBoxSelectTypeEnumMap you had before and just return this in the GetEnumMap function.

scenario/actionhandler.go
```diff
@@ -87,6 +88,19 @@ type (
 	}
 )
 
+type (
+	// *** Interfaces which could be implemented on action stuct field types ***
```
DnlLrssn, line 92: *struct

scenario/actionhandler.go
```diff
@@ -87,6 +88,19 @@ type (
...
+
+	// Enum interface should be implemented on types used fields of action struct if:
+	// 1. Type is derived from one integer type. Example: `type MyType int`
+	// 2. Type has natural string representations for its values.
+	// Typically you should consider implemeting Enum when declaring global
```
DnlLrssn, line 97: * implementing

scenario/listboxselect.go
atluq: I'm with you on that! It will not be optimized just because it could be optimized ;)

PR #76, opened 2020-07-15
Apicompliance/617 Elasticexplore changes

alfinhaji:
**Checklist**

- [ ] tests added
- [x] commits conform to the [Commit message format](./CONTRIBUTING.md#commit)
- [x] documentation updated

**Squash commit message**

Updated both elasticexplore action API from limit=30 to limit=24 to reflect the current statue in QCS.  Also updated the API call change when tag drop down is selected (only seen when the tag parameter is provided in scenario and thus filter by collectionID)

PR #87, opened 2020-07-28
Apicompliance/629 ElasticCreateCollection changes

alfinhaji:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Update GET /api/v1/collections call as part of this action
FROM:  GET /api/v1/collections?query=<collectionName>&type=public&-updateAt
TO:  GET /api/v1/collections?query=<collectionName>&limit=20&sort=%2Bname&type=public

### DIFF
--- a/scenario/elasticexplore.go
+++ b/scenario/elasticexplore.go
@@ -209,7 +209,7 @@ func (settings ElasticExploreSettings) Execute(sessionState *session.State, acti
 
 		// Do we have any tag names
 		for _, collection := range settings.CollectionNames {
-			id, err := searchForTag(sessionState, actionState, host, collection)
+			id, err := searchForTag(sessionState, actionState, host, collection, 100)
 			if err != nil {
 				actionState.AddErrors(errors.WithStack(err))
 				return
@@ -367,13 +367,13 @@ func searchForSpaceByID(sessionState *session.State, actionState *action.State, 
 	}
 }
 
-func searchForTag(sessionState *session.State, actionState *action.State, host, collection string) (string, error) {
+func searchForTag(sessionState *session.State, actionState *action.State, host, collection string, limit int) (string, error) {
 	id, err := sessionState.ArtifactMap.GetStreamID(collection)
 	if err == nil {
 		return id, nil
 	}
 
-	query := fmt.Sprintf("%s/api/v1/collections?query=%s&limit=20&sort=-name&type=public", host, collection)
+	query := fmt.Sprintf("%s/api/v1/collections?query=%s&limit=%d&sort=-name&type=public", host, collection, limit)
 	for {
 		var wg sync.WaitGroup
 		wg.Add(1)
--- a/scenario/elasticexplore.go
+++ b/scenario/elasticexplore.go
@@ -373,7 +373,7 @@ func searchForTag(sessionState *session.State, actionState *action.State, host, 
 		return id, nil
 	}
 
-	query := fmt.Sprintf("%s/api/v1/collections?query=%s&type=public&-updateAt", host, collection)
+	query := fmt.Sprintf("%s/api/v1/collections?query=%s&limit=20&sort=-name&type=public", host, collection)
 	for {
 		var wg sync.WaitGroup
 		wg.Add(1)
--- a/scenario/elasticcreatecollection.go
+++ b/scenario/elasticcreatecollection.go
@@ -45,7 +45,7 @@ func (settings ElasticCreateCollectionSettings) Execute(sessionState *session.St
 		return
 	}
 
-	_, err = searchForTag(sessionState, actionState, host, name)
+	_, err = searchForTag(sessionState, actionState, host, name, 20)
 	if err == nil {
 		actionState.Failed = true
 		sessionState.LogEntry.Log(logger.WarningLevel, "collection already exists")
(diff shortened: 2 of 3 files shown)


alfinhaji:
Tested locally and works 

```
2020-07-28T12:33:02.3761593-04:00	elasticcreatecollection		4	traffic		[31] GET /api/v1/collections?query=Collection-test-aqr2&limit=20&sort=-name&type=public HTTP/1.1\nHost: perf4.us.qlik-stage.com\nUser-Agent: gopherciser \nContent-Type: application/json\nAccept-Encoding: gzip\n\n	Sent		0			SIMPLE_OIDC_USER_Exerciser_94	1	1		79516755201020		0	0		0	0	0	2020-07-28T16:33:02.3761593Z
2020-07-28T12:33:02.4618912-04:00	elasticcreatecollection		4	traffic		[31] HTTP/1.1 200 OK\nContent-Length: 165\nCache-Control: no-store\nConnection: keep-alive\nContent-Type: application/json; charset=utf-8\nDate: Tue, 28 Jul 2020 16:33:39 GMT\nPragma: no-cache\nServer: nginx\nStrict-Transport-Security: max-age=15724800; includeSubDomains\n\n{"data":[],"links":{"self":{"href":"https://perf4.us.qlik-stage.com/api/v1/collections?query=Collection-test-aqr2\u0026limit=20\u0026sort=-name\u0026type=public"}}}\n	Received		0			SIMPLE_OIDC_USER_Exerciser_94	1	1		79516756054217		0	0		0	0	0	2020-07-28T16:33:02.4618912Z
```

scenario/elasticexplore.go
```diff
@@ -373,7 +373,7 @@ func searchForTag(sessionState *session.State, actionState *action.State, host,
 		return id, nil
 	}
 
-	query := fmt.Sprintf("%s/api/v1/collections?query=%s&type=public&-updateAt", host, collection)
+	query := fmt.Sprintf("%s/api/v1/collections?query=%s&limit=20&sort=-name&type=public", host, collection)
```
SeBBBe, line 376: When query gets re-assigned at line 392 (in case of paging) from the links objects in the response, does this query already follow the correct format that client would issue?

SeBBBe: I started checking and realized that might be quite the corner case. Since the parameter query=%s is set to the collection name, this case will only be hit if there are more than 20 collections of the matching name.

SeBBBe:
Another thing I realized though; on line 198 we do the "default fetch" that's supposed to represent the GET of collection that client does upon entering the hub.

`sessionState.Rest.GetAsyncWithCallback(fmt.Sprintf("%s/api/v1/collections?type=public&sort=-name&limit=100", host)....`

Is this still relevant? This line causes the changed code of this PR not to be executed since the collection is already known from the call above. 

I did a quick check just by going to perf4 and checking the traffic and I can only see the client fetching the user's personal collection:

Screenshot 2020-07-29 at 08 32 43

SeBBBe:
Ah I realize now that elasticexplore mimics the behavior of *clicking* the tag dropdown, in which case the first 100 tags are fetched. When I type in a query in the filter pane, it issues a search request with limit=100 though, not 20.

Screenshot 2020-07-29 at 08 35 42

alfinhaji:
> Ah I realize now that elasticexplore mimics the behavior of _clicking_ the tag dropdown, in which case the first 100 tags are fetched. When I type in a query in the filter pane, it issues a search request with limit=100 though, not 20.
> 
> Screenshot 2020-07-29 at 08 35 42

I think what you are referring to (limit=100) when you click the drop down for tags is this limit:  https://github.com/qlik-oss/gopherciser/blob/master/scenario/elasticexplore.go#L198

That was actually updated in this PR:  https://github.com/qlik-oss/gopherciser/pull/76

The 20 limit was observed when creating a new collection (which I could only figure out to do as part of the create app flow)
Create App > start typing in tags section and you will see the limit=20 query

image

alfinhaji:
I see that `elasticUploadApp` also shares this function `SearchForTag`:  https://github.com/qlik-oss/gopherciser/blob/master/scenario/elasticuploadapp.go#L255

So I manually also validated as part of the uploadApp flow when you start typing in the tag field, same limit=20 is seen there as well

image

alfinhaji:
So looks like we are seeing differing behaviour with the limit=100 in the hub during `ElasticExplore` versus `ElasticUploadApp` and `ElasticCreateCollection`

SeBBBe:
Ah, I see. That clarifies things. I'm pushing a commit with a parametrized version of searchForTag so that it's fully compliant in all three cases.

alfinhaji:
Thanks @SeBBBe !  Let me take this branch and validate that it works as expected in all three cases, may just have to wait until next week

alfinhaji:
Just validated and looks good.  Here is `elasticexplore` followed by `elasticcreatecollection`

```
2020-08-10T11:46:30.8121833-04:00	elasticexplore	Open collection	3	traffic		[29] GET /api/v1/collections?type=public&sort=-name&limit=100
2020-08-10T11:46:31.0236321-04:00	elasticcreatecollection		4	traffic		[31] GET /api/v1/collections?query=test.collection&limit=20&sort=-name&type=public 
```

PR #89, opened 2020-08-04
Fixes/630

SeBBBe:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Fixes https://github.com/qlik-trial/gopherciser/issues/619
Fixes https://github.com/qlik-trial/gopherciser/issues/630

### DIFF
--- a/scenario/deletedata.go
+++ b/scenario/deletedata.go
@@ -90,4 +90,10 @@ func (settings DeleteDataSettings) Execute(sessionState *session.State, actionSt
 	if n == 0 {
 		sessionState.LogEntry.Logf(logger.WarningLevel, "no files deleted")
 	}
+
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s/quota", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s?connectionId=&top=1000", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
+	if sessionState.Wait(actionState) {
+		return // we had an error
+	}
 }
--- a/scenario/deletedata.go
+++ b/scenario/deletedata.go
@@ -31,7 +31,10 @@ func (settings DeleteDataSettings) Validate() error {
 }
 
 // Execute action (Implements ActionSettings interface)
-func (settings DeleteDataSettings) Execute(sessionState *session.State, actionState *action.State, connection *connection.ConnectionSettings, label string, reset func()) {
+func (settings DeleteDataSettings) Execute(
+	sessionState *session.State, actionState *action.State, connection *connection.ConnectionSettings, label string,
+	reset func(),
+) {
 	host, err := connection.GetRestUrl()
 	if err != nil {
 		actionState.AddErrors(err)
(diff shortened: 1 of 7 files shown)


scenario/deletedata.go
```diff
@@ -90,4 +90,10 @@ func (settings DeleteDataSettings) Execute(sessionState *session.State, actionSt
 	if n == 0 {
 		sessionState.LogEntry.Logf(logger.WarningLevel, "no files deleted")
 	}
+
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s/quota", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s?connectionId=&top=1000", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
```
atluq, line 95: No connectionid?

scenario/uploaddata.go
```diff
@@ -122,4 +124,10 @@ func (settings UploadDataSettings) Execute(sessionState *session.State, actionSt
 		actionState.AddErrors(errors.New(fmt.Sprintf("failed to upload data file payload: %d <%s>", postData.ResponseStatusCode, postData.ResponseBody)))
 		return
 	}
+
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s/quota", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
+	sessionState.Rest.GetAsync(fmt.Sprintf("%s/%s?connectionId=&top=1000", host, datafileEndpoint), actionState, sessionState.LogEntry, nil)
```
atluq, line 129: No connectionid?

SeBBBe:
Good catch! According to Alfin in the issue: https://github.com/qlik-trial/gopherciser/issues/619 connectionId should be empty. When I tried it, I could see that connectionId wasn't empty, but I also don't know where the id comes from so I figured I'd leave it like it was specified in the issue and we can have a discussion around it :)

alfinhaji:
Sorry must have meant to replace the actual connection ID when I copied and pasted with `<connectionID>` when I documented this in https://github.com/qlik-trial/gopherciser/issues/619, but yes there is a connectionID, and I suspect its the same as what  I implemented here as part of elasticopenhub:  https://github.com/qlik-oss/gopherciser/pull/75

Connection ID is obtained during elasticopenhub, and stored as `qid`: https://github.com/qlik-oss/gopherciser/blob/master/scenario/elasticopenhub.go#L230

Its part of the response from this: https://github.com/qlik-oss/gopherciser/blob/master/scenario/elasticopenhub.go#L211

SeBBBe:
Oh, all right. Will look at saving that value for later together with the app and then re-using it for these actions.

SeBBBe:
Do you know what the point of the connectionId is? Is it supposed to be tenant-specific?

SeBBBe:
Added

`DataConnectionId string`

in the SessionState struct. It's saved whenever an openhub is executed, and used in the data manipulation actions. If no openhub preceeds the action, an empty connectionId will be used (seems to work fine at least functionally)

alfinhaji:
> Do you know what the point of the connectionId is? Is it supposed to be tenant-specific?

I dont know the answer to this unfortunately.

PR #90, opened 2020-08-04
Enhancement/action/clickactionbutton

atluq:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

Add action-button support

Adds:
* Clickactionbutton-action
* Support for caching of fields and variables
* Helper functions for more concise json marshal and unmarshal of enum type
* Helper functions for caching fields and variables
* Sorted sheetlist

**Info**
Solves #607


scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+)
+
+// IsContainerAction implements ContainerAction interface
+// and sets container action logging to original action entry
+func (settings ClickActionButtonSettings) IsContainerAction() {}
+
+// Validate filter pane select action
+func (settings ClickActionButtonSettings) Validate() error {
+	if settings.ID == "" {
+		return errors.Errorf("Empty object ID")
```
SeBBBe, line 104: Would prefer to keep the error messages all lower-case as mentioned above (the current implementations are probably kinda mixed, we should go back and change..).

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+		}
+
+		actions := []buttonAction{}
+		err = json.Unmarshal(actionsRaw, &actions)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "Failed to unmarshal button actions")
+		}
+
+		// parse navigation action associated with button
+		navigationRaw, err := senseobjdef.NewDataPath("navigation").Lookup(propsRaw)
```
SeBBBe, line 181: Do both actions and navigation exist for all buttons? So there's no buttons with exclusively one or the other?

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+
+	switch buttonAction.ActionType {
+	case emptyAction: // do nothing
+		return nil
+
+	case unknownAction:
+		return errors.New("Unknown button action")
+
+	case applyBookmark:
+		return sendReq(func(ctx context.Context) error {
```
SeBBBe, line 215: You should call the existing ApplyBookmark action here instead.

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+		})
+
+	case moveBackwardsInSelections:
+		return sendReq(doc.Back)
+
+	case moveForwardsInSelections:
+		return sendReq(doc.Forward)
+
+	case clearAllSelections:
+		return sendReq(func(ctx context.Context) error {
```
SeBBBe, line 233: This could also be implemented with the existing Select action

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+			field, err := uplink.FieldCache.LookupWithFallback(buttonAction.Field,
+				func(fieldName string) (*enigma.Field, error) {
+					return doc.GetField(ctx, fieldName)
+				},
+			)
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			_, err = field.ClearAllButThis(ctx, buttonAction.SoftLock)
```
SeBBBe, line 247: That's interesting! Maybe that would be a useful option to add to the ones in the Select action.

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			_, err = field.ToggleSelect(ctx, buttonAction.Value /*match*/, buttonAction.SoftLock, 0 /*excludedValuesMode*/)
+			return errors.Wrapf(err, "Could not select possible in field<%s>", buttonAction.Field)
+		})
+
+	case lockAllSelections:
+		return sendReq(func(ctx context.Context) error {
+			return doc.LockAll(ctx, "" /*stateName*/)
```
SeBBBe, line 367: Might be interesting to keep track of this internally so that we can send an appropriate warning if the user later does something which is not allowed under the current state.

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+	case nextSheet:
+		currentSheet, err := GetCurrentSheet(sessionState.Connection.Sense())
+		if err != nil {
+			return errors.Wrapf(err, "Could not get current sheet")
+		}
+		currentSheetIdx, ok := IndexOf(currentSheet.ID, sheets)
+		if !ok {
+			return errors.Wrapf(err, "Could not get current sheet")
+		}
+		nextSheetIdx := (currentSheetIdx + 1) % len(sheets)
```
SeBBBe, line 459: 👍

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,643 @@
...
+			return errors.New("Empty sheet id")
+		}
+		sheetID = navAction.Sheet
+
+	default:
+		return errors.Errorf("Button navigation action '%s' is not supported", navAction.Action)
+	}
+
+	// Execute changeSheet Action
+	changeSheetAction := Action{
```
SeBBBe, line 489: I think this is a great way to structure the navigation.

scenario/helpers.go
```diff
@@ -95,3 +100,83 @@ func Contains(list []string, match func(s string) bool) bool {
...
+		if str == match {
+			return i, true
+		}
+	}
+	return -1, false
+}
+
+// TODO move enum related code to its own package
+type (
+	MutableEnum interface {
```
SeBBBe, line 116: Why do we need these new types?

atluq:
Nice feedback!

I will reuse the existing actions for selection and applybookmark. 

Field-selections do not exist in gopherciser yet, but I will take on implementing it now when up to speed with the field-api (after evaluting what is possible in the client).

I will give a second thought on duplication in the cache classes. Feel little bit handcuffed by the type system ;)

docs/settingup.md
```diff
@@ -2042,6 +2042,29 @@ Upload a data file to the Data manager.
 ```
 
 </details>
+</details>
+<details>
+<summary>Ungrouped actions</summary>
```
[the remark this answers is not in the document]
atluq: Missed that, will fix 👍

scenario/applybookmark.go
```diff
@@ -35,7 +36,7 @@ func (settings ApplyBookmarkSettings) Validate() error {
 // Execute ApplyBookmarkSettings action (Implements ActionSettings interface)
 func (settings ApplyBookmarkSettings) Execute(sessionState *session.State, actionState *action.State, connectionSettings *connection.ConnectionSettings, label string, reset func()) {
 	if sessionState.Connection == nil || sessionState.Connection.Sense() == nil {
-		actionState.AddErrors(errors.New("not connected to a Sense environment"))
+		actionState.AddErrors(errors.New("Not connected to a Sense environment"))
```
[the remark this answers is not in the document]
atluq: Will change that in every place. This one must have accidentally ended up in this commit.

SeBBBe:
> I will give a second thought on duplication in the cache classes. Feel little bit handcuffed by the type system ;)
Yup I know that feeling 😆 

Keep in mind that everything the user can do via actions are supposed to be representations of something a real user could actually do using the client. I.e I'm not sure how to do a ClearAllButThis manually using the regular Sense client, but if it's possible it'd be neat to add it to Select.

scenario/helpers.go
atluq: Fewer lines of boilerplate when implementing json.Marshaler and json.Unmarshaler for Enum-types. Had to do that for two Enum-types at the end of clickactionbutton.go.

scenario/clickactionbutton.go
atluq: Very true. Will take a look at where.

scenario/clickactionbutton.go
atluq: There exist buttons with one or the other, but the fields are never omitted in messages. There is a "none" navigation action and an empty list for the other actions.

enigmahandlers/fieldcache.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (fc *fieldCache) Lookup(name string) (field *enigma.Field, hit bool) {
+	fc.mutex.RLock()
+	defer fc.mutex.RUnlock()
+	field, hit = fc.fieldMap[name]
+	return
+}
+
+func (fc *fieldCache) LookupWithFallback(fieldName string, fallback func(fieldName string) (*enigma.Field, error)) (*enigma.Field, error) {
```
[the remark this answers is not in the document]
atluq: Its meant as some kind of caching function decorator.

It does not always have to fetch, if I understand your question.

The behaviour I am after is: take from cache, or else, fetch and store.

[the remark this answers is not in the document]
SeBBBe: Exactly. So the fetch logic could just be in the function rather than sending it in each time. Because right now, the following part:

`func(fieldName string) (*enigma.Field, error) {
					return doc.GetField(ctx, fieldName)
				},`

is duplicated for each function invocation. So instead of a "LookupWithFallback" function more like a "LookupOrFetch"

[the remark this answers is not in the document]
atluq: Cool! I see what you mean. I will write such functions to reduce duplication. They will end up outside of the caching classes though.

scenario/clickactionbutton.go
```diff
@@ -196,109 +196,117 @@ func buttonActions(sessionState *session.State, actionState *action.State, obj *
...
-		})
-
 	case clearSelectionsInOtherFields:
 		return sendReq(func(ctx context.Context) error {
-			field, err := uplink.FieldCache.LookupWithFallback(buttonAction.Field,
-				func(fieldName string) (*enigma.Field, error) {
-					return doc.GetField(ctx, fieldName)
-				},
-			)
+			field, err := fieldReq(doc.GetField).WithCache(&uplink.FieldCache).WithInputValidation()(ctx, buttonAction.Field, "")
```
SeBBBe, line 261: We haven't used much function chaining in the project before so something to be aware of. But the end result looks elegant in this use case. Nicely done!

enigmahandlers/fieldcache.go
```diff
@@ -7,38 +7,26 @@ import (
...
 
-func (fc *fieldCache) Lookup(name string) (field *enigma.Field, hit bool) {
+func (fc *FieldCache) Lookup(name string) (field *enigma.Field, hit bool) {
 	fc.mutex.RLock()
 	defer fc.mutex.RUnlock()
 	field, hit = fc.fieldMap[name]
 	return
 }
 
-func (fc *fieldCache) LookupWithFallback(fieldName string, fallback func(fieldName string) (*enigma.Field, error)) (*enigma.Field, error) {
```
SeBBBe: Always  good to see duplication go away :) Nice solution

scenario/clickactionbutton.go
```diff
@@ -0,0 +1,629 @@
...
+func (buttonAction *buttonAction) execute(sessionState *session.State, actionState *action.State,
+	connectionSettings *connection.ConnectionSettings, label string) error {
+
+	uplink := sessionState.Connection.Sense()
+	doc := uplink.CurrentApp.Doc
+	sendReq := func(f func(context.Context) error) error {
+		return sessionState.SendRequest(actionState, f)
+	}
+	executeContainerAction := func(actionType string, settings ActionSettings) error {
+		return executeContainerAction(sessionState, connectionSettings, actionType, settings)
```
SeBBBe, line 227: Nice way to cut down on the number of arguments!

scenario/clickactionbutton.go
atluq: Thanks!

The enigma.Doc has a lot of functions that will be wrapped with the same logic in a lot of places. It Doc only were an interface, we could more easy wrap that one and reuse it in many places.

atluq:
> I will approve the PR. Next step, however, is to have it tested before merging. @alfinhaji @ashishkumarqlik any of you volunteer? :) Pay extra attention to the logging and analysis as we haven't used container actions that much in the past.

Yes, an actionbutton contain a list of button-actions and one optional navigation action. The types which are implemented as container actions are:
- All types of navigation actions (mostly change sheet actions)
- ApplyBookMark
- ClearAllSelections

So keep an extra eye on those if you take the time to test this.

> Another concern is whether any of the actions might have side effects on the rest of the scenario following this action that we need to handle somehow. I already know that the app structure will be invalid, but that's also the case for RandomAction and some other actions that we can't statically analyze. Perhaps we will discuss further how to handle these cases in the future.

Yes, it would be nice with a discussion on how to keep the state in the best way, such that the user will know when actions do not have the supposed effect. I created an issue on "locked selections" which falls into this discussion.

generatedocs/data/actions/clickactionbutton/description.md
```diff
@@ -0,0 +1,4 @@
+## ClickActionButton action
+
+Simulate clicking an action-button
```
carlcervin, line 3: Maybe change to:
"Simulate clicking an action-button."
By the way, do we need an explanation of what an "action-button" is?

generatedocs/data/actions/clickactionbutton/examples.md
```diff
@@ -0,0 +1,11 @@
+### Examples
+
+```json
+{
+     "label": "ClickActionButton",
+     "action": "ClickActionButton",
+     "settings": {
+         "id": "951e2eee-ad49-4f6a-bdfe-e9e3dddeb2cd",
```
carlcervin, line 8: Remove the comma at the end?

generatedocs/data/actions/clickactionbutton/description.md
SeBBBe: Yes, I think having a brief explanation here of the supported actions of an action button might be useful.

docs/settingup.md
```diff
@@ -436,9 +436,34 @@ Clear all selections in an app.
 
 ## ClickActionButton action
 
-Simulate clicking an action-button
-
-
+A `ClickActionButton`-action simulates clicking an _action-button_. An _action-button_ is a chart item which, when clicked, executes a series of actions. The series of actions contained by an action-button begins with any number _generic button-actions_ and ends with an optional _navigation button-action_.
```
SeBBBe, line 439: Chart item? Sheet item, perhaps?

docs/settingup.md
```diff
@@ -436,9 +436,34 @@ Clear all selections in an app.
...
+
+**Note:** Clicking an action-button may have side effects such as changing sheet and locking selections, which highly affect the outcome of following actions.
```
SeBBBe, line 441: I'm not sure if we need this disclaimer. Many of the actions affect the state of the app and thus any succeeding actions, but we don't have any such note on those. I think this may mainly be an internal concern for the app structure functionality since it leaves the app in an undetermined state.

docs/settingup.md
atluq: Fixing, using that term is just confusing. It is a chart in some meaning though 😄 
image

docs/settingup.md
atluq: This could surely be noted in most actions, but since the button contain a "hidden" macro, the script-authors may have some extra need of a heads up.

atluq: But as you say, the app-state should be kept to give appropriate warnings :)

DnlLrssn:
Also has the reporting been verified to look correctly?

atluq:
> Also has the reporting been verified to look correctly?

It is not verified, Monica shall test this.

I guess this has to be tested extra carefully due to the use of container actions.

DnlLrssn:
> I guess this has to be tested extra carefully due to the use of container actions.

Yes, that's my thought, just some extra carefulness due to added complexity . Should also be verify with "developer eyes" in addition to testers imho.

atluq:
> a bit unsure about our own cache for fields and variables, haven't checked, but there's no FieldList and/or VariableList sessionobject that the client uses or something? With a session object adding and removing of fields (is dynamic fields still a thing?) and variables would be handled (as long as events are set up).

Fields and variables are cached by the client, but this implementation may be too much of a simplification. We should discuss this further...

atluq:
> > I guess this has to be tested extra carefully due to the use of container actions.
> 
> Yes, that's my thought, just some extra carefulness due to added complexity . Should also be verify with "developer eyes" in addition to testers imho.

The developer reason is to not repeat ourselves. The button action though is not a container action in the same sense as the other container actions, because the container part is defined by sense and not by gopherciser.

If button-action shall not be logged as a container action, we should find a way to reuse actions without the container action logging.

DnlLrssn:
@Mounikabanka when validating, please be extra careful to validate results, how it shows in analyzer responsetime reporting etc

SeBBBe (approved):
The new code since last time LGTM!

atluq:
Propagate label to subactions!

atluq:
> Propagate label to subactions!

Fixed.

scenario/clickactionbutton.go
```diff
@@ -133,23 +133,36 @@ func (settings ClickActionButtonSettings) Execute(sessionState *session.State, a
 		return
 	}
 
+	label = firstNonEmpty(label, "buttonaction")
+
 	// run button-actions
 	for _, buttonAction := range buttonActions {
-		err := buttonAction.execute(sessionState, actionState, connectionSettings, label)
+		buttonActionLabel := fmt.Sprintf("%s: %s", label, firstNonEmpty(buttonAction.ActionLabel, buttonAction.ActionType.String()))
```
DnlLrssn, line 140: when using %s the stringer interface will be used automatically, i.e. .String() is unnecessary

atluq: `buttonAction.ActionType.String()` is an argument to `firstNonEmpty`

DnlLrssn: ah, you're right, sorry

scenario/helpers.go
DnlLrssn: What's the big difference from existing
func (em *EnumMap) UnMarshal(arg []byte) (int, error) {...}
func (em *EnumMap) StringDefault(i int, dflt string) string {...}

DnlLrssn:
re-posting comment since it's gotten kind-of hidden

https://github.com/qlik-oss/gopherciser/pull/90/files#r481101620
> What's the big difference from existing
> func (em *EnumMap) UnMarshal(arg []byte) (int, error) {...}
> func (em *EnumMap) StringDefault(i int, dflt string) string {...}

atluq:
> re-posting comment since it's gotten kind-of hidden
> 
> https://github.com/qlik-oss/gopherciser/pull/90/files#r481101620
> 
> > What's the big difference from existing
> > func (em *EnumMap) UnMarshal(arg []byte) (int, error) {...}
> > func (em *EnumMap) StringDefault(i int, dflt string) string {...}

Great you did. Would have missed it.

Yeah, it is not so good to have two versions of utilities for this.

PR #96, opened 2020-08-21
Make qID fetch in elasticopenhub optional

SeBBBe:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

If no connection ID is returned in ElasticOpenHub, throw a warning rather than an error. Instead, retry the fetch when doing an upload or deleting a file (client does this upon entering DLE). An error will still be thrown if it fails at this point.

### DIFF
--- a/scenario/deletedata.go
+++ b/scenario/deletedata.go
@@ -48,7 +48,7 @@ func (settings DeleteDataSettings) Execute(
 	restHandler := sessionState.Rest
 
 	if sessionState.DataConnectionId == "" {
-		FetchQid(sessionState, actionState, host, true)
+		FetchDataConnectionId(sessionState, actionState, host, true)
 		if sessionState.Wait(actionState) {
 			return // we had an error
 		}
--- a/scenario/deletedata.go
+++ b/scenario/deletedata.go
@@ -47,6 +47,13 @@ func (settings DeleteDataSettings) Execute(
 
 	restHandler := sessionState.Rest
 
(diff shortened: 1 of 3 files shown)


DnlLrssn:
> 
> 
> qid fetch is now shown as a warning. Tested with reconnect true and false. Looks good.

Does this include tests with actions actually handling datafiles?

alfinhaji:
> approving the code, but still not certain this is the behavior we want, @alfinhaji ?

My understanding of the issue is that when the system is under heavy load and dependant services are stressed, the qid can sometimes not be returned.  I think this mainly manifests during Freddy's tests.  
https://qlikdev.slack.com/archives/G01597JLP7W/p1598016282041000

DnlLrssn:
> 
> My understanding of the issue is that when the system is under heavy load and dependant services are stressed, the qid can sometimes not be returned. I think this mainly manifests during Freddy's tests.
> https://qlikdev.slack.com/archives/G01597JLP7W/p1598016282041000

In which case I would consider it an error from performance and stability perspective.

foh79:
Issue occurs during heavy load yes. Not seen with fewer users. 
Also tested some uploading and datafiles and stuff. Did not see any problems there.

So what do we do? What did the owners of the service say Fabian? Where there any plans on fixing it?

SeBBBe:
I filed a bug on the behavior: https://jira.qlikdev.com/browse/QB-2166

Seems they're planning on returning an actual error code when things go wrong.

PR #101, opened 2020-09-04
Replace `tick` preciontime with atomic counter

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Replace `tick` log entry with straight counter and remove preciontime package. Tick and preciontime only used for logging and mainly for sort order nowadays, replacing with straight counter fills same purpose for current usage.

**Info**
old tick:
image

new tick:
image

Does not affect CPU or memory:

image

But improves syscall time:
image

Which means less blocking of goroutines.


SeBBBe (approved):
Nice! This is the kind of thing we should go back and do more often.

Also:
Screenshot 2020-09-07 at 09 13 58

PR #106, opened 2020-09-11
Enhancement/644 add story objects to app structure

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add new array in structure array `storyobjects`. 
* Move all story and snapshot type of objects to `storyobjects` array instead of `objects` array.
* Connect all stories and snapshots so hierarchy can be traced. 
* Avoid getting app structure failing due to having `embeddedsnapshot" objects.

**Info**

Example file ( renamed to .txt so github would let me add it).
[31e1ab83-c896-4252-920f-a428ae5cbafe.txt](https://github.com/qlik-oss/gopherciser/files/5208876/31e1ab83-c896-4252-920f-a428ae5cbafe.txt)

when using the `--raw` flag, these fields will be added for some objects:
- `rawProperties`
- `rawSnapshotProperties`

### DIFF
--- a/config/appstructure.go
+++ b/config/appstructure.go
@@ -393,7 +393,7 @@ func extractGeneratedProperties(properties json.RawMessage) json.RawMessage {
 	return properties
 }
 
-func handleChildren(ctx context.Context, genObj *enigma.GenericObject, obj *appstructure.AppStructureObjectChildren) error {
+func handleChildren(ctx context.Context, genObj *enigma.GenericObject, children *appstructure.AppStructureObjectChildren) error {
 	childInfos, err := genObj.GetChildInfos(ctx)
 	if err != nil {
 		return errors.WithStack(err)
@@ -403,10 +403,10 @@ func handleChildren(ctx context.Context, genObj *enigma.GenericObject, obj *apps
 		if child == nil {
 			continue
 		}
-		if obj.Children == nil {
-			obj.Children = make(map[string]string)
+		if children.Map == nil {
+			children.Map = make(map[string]string)
 		}
-		obj.Children[child.Id] = child.Type
+		children.Map[child.Id] = child.Type
 	}
 
 	return nil
--- a/config/appstructure.go
+++ b/config/appstructure.go
@@ -269,7 +269,8 @@ func (structure *GeneratedAppStructure) getStructureForObjectAsync(sessionState 
 			fallthrough
 		case appstructure.ObjectStory, appstructure.ObjectSlide, appstructure.ObjectImage, appstructure.ObjectText,
 			appstructure.ObjectShape, appstructure.ObjectSlideItem, appstructure.ObjectSnapshotList:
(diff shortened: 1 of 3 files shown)


config/appstructure.go
```diff
@@ -138,6 +138,9 @@ func (structure *GeneratedAppStructure) printSummary(summary SummaryType, fileNa
 func evaluateActionList(actions []scenario.Action, includeRaw bool, summary SummaryType) []scenario.Action {
 	appStructureScenario := make([]scenario.Action, 0, len(actions))
 	for _, act := range actions {
+		if act.Disabled {
```
SeBBBe, line 141: Good catch!

SeBBBe (approved):
Aside from code review also tested this build in QSEST, existing functionality seems to still work fine.

config/appstructure.go
DnlLrssn: yes somehow forgot about that the first time around

atluq (commented):
Nice we are adding story objects to app structure!

One thing I noticed is that `AppstructureObjectsChildren.Children` stutters a bit.

DnlLrssn:
> 
> 
> One thing I noticed is that `AppstructureObjectsChildren.Children` stutters a bit.

Hmm true, 
We don't want to change the json tag from children,  and preferably not  inner field so it corresponds to tag, but the struct name could be changed. Can't think of good though and has to be unique. Also don't want to use "childlist" since that's something that exist and is something different.

atluq:
> Hmm true,
> We don't want to change the json tag from children, and preferably not inner field so it corresponds to tag, but the struct name could be changed. Can't think of good though and has to be unique. Also don't want to use "childlist" since that's something that exist and is something different.
> 
> so any suggestions?

Cant find a good name for the struct. Just having a map named and tagged "children" as a field in every struct where it is needed. Pretty much like before 😄  I dont think this kind of duplication is bad. And change the parameter type to map if we dont want to pass the whole object.

DnlLrssn:
Added it so this part can be passed to function which also makes the map if nil, independent of type of mother object
 
```golang
if err := handleChildren(ctx, extendedObject, &obj.AppStructureObjectChildren); err != nil {
```

DnlLrssn:
if it's of type map it is either nil, and doesn't have an pointer address , or it needs to be allocated before call = will always be not nil = will not be omitted in json.

DnlLrssn:
i.e. re-used existing code, since didn't want to duplicte code when adding same kind of logic to a different motherobject.

atluq (approved):
Now I see why it is a separate struct 👍

DnlLrssn:
> Now I see why it is a separate struct +1

But now the name annoys me though, maybe I think of something until Monday...

DnlLrssn:
Didn't come up with a good name for the struct, so renamed the field anyway, but of course kept the json tag name as is.

PR #110, opened 2020-09-21
Add possibility to continue on errors in scheduler

SeBBBe:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Adds a possibility to call a function on the scheduler to continue on errors.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -51,6 +51,8 @@ type (
 		ReconnectSettings session.ReconnectSettings `json:"reconnectsettings" doc-key:"config.scheduler.reconnectsettings"`
 
 		connectionSettings *connection.ConnectionSettings
+
+		continueOnErrors bool
 	}
 
 	schedulerTmp struct {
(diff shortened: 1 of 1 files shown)


scheduler/scheduler.go
```diff
@@ -235,7 +246,9 @@ func (sched *Scheduler) runIteration(userScenario []scenario.Action, sessionStat
 				logEntry.Action = nil
 				logEntry.LogError(errors.Wrap(errTimeBuf, "time buffer in-between sequences failed"))
 			}
-			return errors.WithStack(err)
+			if !sched.continueOnErrors {
+				return errors.WithStack(err)
```
DnlLrssn, line 250: Maybe log warning since otherwise error won't get reported? Alternative 2, collect all errors in a multierror and return at end.

SeBBBe: Ah really? I figured it would be added to the logger either way.

SeBBBe: As far as I can tell, it would only affect the return of the Execute method, logging should still be the same, no? I still want to log the errors as normal.

DnlLrssn (approved):

PR #111, opened 2020-09-23
Enhancement/deprecate elastic reload parameters

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

* Update elasticreload to deprecate polling flags:
    * Deprecate `pollinterval`.
    * Deprecate `pollingoff.`
* Update documentation.
* Add deprecated fields helper for easier and more generic usage.
* ReloadDuration logging will not be re-added after all, remove commented code.

**Info**

```bash
JSONParseError: Failed to unmarshal config from json: unmarshalerDecoder: Failed unmarshaling config: config.cfgCore.Scenario: []scenario.Action: unmarshalerDecoder: Failed to unmarshal action<elasticreload> settings: unmarshalerDecoder: elasticreload has deprecated fields: /pollingoff, please remove from script, error found in #10 byte of ...
```


scenario/elasticreload.go
```diff
@@ -19,27 +18,13 @@ import (
 
 type (
 	// ElasticReloadCore Currently used ElasticReloadCore (as opposed to deprecated settings)
-	ElasticReloadCore struct {
-		// PollInterval time in-between polling for reload status
-		PollInterval helpers.TimeDuration `json:"pollinterval" displayname:"Poll interval" doc-key:"elasticreload.pollinterval"`
-		// PollingOff turns polling for status off except after event websocket reconnect
-		PollingOff bool `json:"pollingoff" displayname:"Polling off" doc-key:"elasticreload.pollingoff"`
-	}
+	ElasticReloadCore struct{}
```
SeBBBe, line 21: Is there a reason to keep this empty struct around?

scenario/helpers.go
```diff
@@ -140,3 +143,18 @@ func (getVar varReq) WithCache(vc *enigmahandlers.VarCache) varReq {
 		return variable, nil
 	}
 }
+
+// HasDeprecatedFields check json if keys exist at provided paths, if any exist report error
+func HasDeprecatedFields(rawJson []byte, deprecatedPaths []string) error {
```
SeBBBe, line 148: 👍

atluq: Useful!

atluq (commented):
No polling 🥳 

Consider removing empty ElasticReloadCore as Fabian noted.

scenario/elasticreload.go
DnlLrssn: hmm not directly. Might be good to live it in case we want to add other parameters in the future since we already have the duouble core things working here (considering the unmarshal bug if you don't do it this way). Other than that it could actually be removed.

SeBBBe: If it solves a problem it's fine. We shouldn't keep it around just because we might want it in the future - we already have it in the git history.

SeBBBe (approved):

scenario/elasticreload.go
DnlLrssn: not because we "might need it" more because it's easy to think that adding a field to the main struct will work, but it won't because of the bug in json.Unmarshal, so will remind of that saving hours of debugging if we forget. But sure, wasn't directly intentional to leave it, just figured now that maybe it was a good idea, but if no-one agrees I'll remove it.

SeBBBe: Ah right. Now I see what you mean.

PR #116, opened 2020-10-01
Fix panic when no connection during reload

atluq:
**Checklist**

**Squash commit message**

Error instead of panic if no connection during reload

**Info**

Closes #675

### DIFF
--- a/scenario/reload.go
+++ b/scenario/reload.go
@@ -73,11 +73,16 @@ func (settings ReloadSettings) Validate() error {
 func (settings ReloadSettings) Execute(sessionState *session.State,
 	actionState *action.State, connectionSettings *connection.ConnectionSettings, label string, reset func()) {
 
-	if sessionState.Connection == nil || sessionState.Connection.Sense() == nil {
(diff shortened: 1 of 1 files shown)


DnlLrssn (approved):

DnlLrssn:
hmm wait, maybe we should check connection too?

atluq:
Yes, let's do that.

scenario/reload.go
```diff
@@ -101,8 +106,12 @@ func (settings ReloadSettings) Execute(sessionState *session.State,
 				// Get the progress using the request id we reserved for the reload
 				var progress *enigma.ProgressData
 				getProgress := func(ctx context.Context) error {
+					uplink := sessionState.Connection.Sense()
```
DnlLrssn, line 109: still should check sessionState.Connection != here as this is in separate go-routine and something else could nil the connection.

atluq: ahh, then getting an uplink from sessionState should be atomic or we should just accept the panics

DnlLrssn (approved):

PR #117, opened 2020-10-01
Fix an issue where some linecharts yield a hypercube error

SeBBBe:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

Fix an issue where some linecharts yield a hypercube error

**Info**

Change lookup for whether a linechart is considered continuous, and therefore needs to be fetched with GetHyperCubeContinuousData.

Change constraint to be an array, where each constraint needs to be fulfilled. Throw an unmarshal error if an object definition file of the old format is used.

### DIFF
--- a/cmd/objdef.go
+++ b/cmd/objdef.go
@@ -212,10 +212,10 @@ func fileExists(name string) bool {
 
 func constraintString(data senseobjdef.Data, buf *helpers.Buffer) string {
 	if data.Constraints == nil {
-		buf.WriteString("|   Constraints: Default\n")
+		buf.WriteString("|   Constraint: Default\n")
 	} else {
 		for _, c := range data.Constraints {
-			buf.WriteString("|   Constraints: ")
+			buf.WriteString("|   Constraint: ")
 			buf.WriteString("[")
 			buf.WriteString(string(c.Path))
 			buf.WriteString("] ")
--- a/cmd/objdef.go
+++ b/cmd/objdef.go
@@ -211,11 +211,11 @@ func fileExists(name string) bool {
 }
 
 func constraintString(data senseobjdef.Data, buf *helpers.Buffer) string {
-	if data.Constraint == nil {
-		buf.WriteString("|   Constraint: Default\n")
+	if data.Constraints == nil {
+		buf.WriteString("|   Constraints: Default\n")
 	} else {
-		for _, c := range data.Constraint {
-			buf.WriteString("|   Constraint: ")
+		for _, c := range data.Constraints {
+			buf.WriteString("|   Constraints: ")
 			buf.WriteString("[")
 			buf.WriteString(string(c.Path))
 			buf.WriteString("] ")
--- a/cmd/objdef.go
+++ b/cmd/objdef.go
@@ -95,8 +95,8 @@ Use to export default values or examples or to validate custom definitions or ov
 		Use:     "validate",
 		Aliases: []string{"val", "v"},
 		Short:   "Validate object definitions in file.",
-		Long: `Validate object definitions from provided JSON file. Will print how many definitions where found,
-it's recommended to use to -v verbose flag and verify all parameters where interpreted correctly'`,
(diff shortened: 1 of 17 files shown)


senseobjdef/constraint.go
```diff
@@ -138,7 +142,12 @@ func (constraint *Constraint) Evaluate(data json.RawMessage) (bool, error) {
 	case string:
 		return constraint.operator.evalString(value.(string), constraint.value)
 	default:
-		return false, errors.Errorf("value type<%T> not supported", value)
+		switch reflect.TypeOf(value).Kind() {
+		case reflect.Slice:
+			return constraint.operator.evalArray(value, constraint.value)
+		default:
+			return false, errors.Errorf("value type<%T> not supported", value)
+		}
```
atluq, line 150: Do we expect more of the slice type here?

It might be overkill to use reflection instead of a fourth `case []string, []float64, []int` or whatever is appropriate.

senseobjdef/constraint.go
```diff
@@ -192,3 +201,29 @@ func (operator constraintOperator) evalString(val string, constraint string) (bo
...
+
+func (operator constraintOperator) evalArray(val interface{}, constraint string) (bool, error) {
+	v := reflect.ValueOf(val)
+	n := v.Len()
+
+	switch operator {
+	case containsOperator:
+		for i := 0; i < n; i++ {
+			e := v.Index(i).Interface()
+			str, ok := e.(string)
```
DnlLrssn, line 213: should maybe support ints and float arrays too? i.e. a switch on type then cast

DnlLrssn (commented):
:+ Datapath belongs a lot better as helper as we started using it all over
:- Not superhappy with reflection in performance sensitive part of the code. Datapath is already a bit too slow as we've seen before.
? Has object defintions override from file been tested?

atluq (commented):
Looks good 👍 

Would prefer if someone else also took a look before approve.

The use of reflect is possibly not necessary (see inline comment).

SeBBBe:
:+ Datapath belongs a lot better as helper as we started using it all over
Thought so too!

:- Not superhappy with reflection in performance sensitive part of the code. Datapath is already a bit too slow as we've seen before.
I'll remake it without reflection! Good point actually.

? Has object defintions override from file been tested?
Yes, it was somewhat painstakingly tested as I had to fight with some pointers 😒  But it works now!

senseobjdef/constraint.go
SeBBBe: You're right, that would be better!

senseobjdef/constraint.go
SeBBBe: Sure, why not

cmd/objdef.go
```diff
@@ -95,8 +95,8 @@ Use to export default values or examples or to validate custom definitions or ov
 		Use:     "validate",
 		Aliases: []string{"val", "v"},
 		Short:   "Validate object definitions in file.",
-		Long: `Validate object definitions from provided JSON file. Will print how many definitions where found,
-it's recommended to use to -v verbose flag and verify all parameters where interpreted correctly'`,
+		Long: `Validate object definitions from provided JSON file. Will print how many definitions were found,
+it's recommended to use to -v verbose flag and verify all parameters were interpreted correctly'`,
```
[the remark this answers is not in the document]
SeBBBe: I also did some search + replace on "contraint" 😆  I've made that mistake myself too though..

DnlLrssn (approved):

DnlLrssn:
@QlikFrederic This is a breaking change for overriding/adding object definitions from file. Does benchmarking package still use that? If so will need updating to new structure after this has been merged and included in a release.

atluq (approved):

SeBBBe:
To Frederic: It's a very simple change though, and there's a check that will throw an error in case the old format is used. Just change "constraint" to "constraints" and add a pair of [] to make it into an array rather than a single object.

PR #120, opened 2020-10-06
Enhancement/extensible generatedocs

atluq:
**Checklist**

- [x] tests added
- [] documentation updated

**Squash commit message**

Add functionality to override generatedocs data.

**Info**

Will update docs when we settle for a solution.


generatedocs/compile/compile.go
```diff
@@ -69,10 +104,100 @@ func handleFlags() {
...
+	mergedGroups := make([]common.GroupsEntry, 0, len(baseGroups)+len(newGroups))
+
+	// merge groups existing in base
+	for _, baseGroup := range baseGroups {
+		if newGroup, existInBase := newGroupMap[baseGroup.Name]; existInBase {
+			// mark new group as merged by deleting it from lookup table
+			delete(newGroupMap, baseGroup.Name)
+
+			// override string fields
+			if baseGroup.Description != "" {
```
SeBBBe, line 143: Why overwrite only when it's non-empty?

generatedocs/compile/testdata/base/data/actions/changesheet/description.md
```diff
@@ -0,0 +1,25 @@
...
+* Pie chart
+* Tree map
+* Text-Image
+* KPI
+* Gauge
+* Box plot
+* Distribution plot
+* Histogram
+* Auto chart (including any support generated visualization from this list)
+* Waterfall chart
```
SeBBBe, line 25: Not related to this PR, but this information does not seem to belong here IMO. It doesn't seem justified to have an extra listing of every single object for this particular action.

SeBBBe:
Shouldn't the docs in the original location (generatedocs/data/actions) be removed?

generatedocs/compile/compile.go
atluq: No reason 🙈  it is supposed to be newGroup. Will make sure this is tested to, the tests are only with empty Descriptions etc.

SeBBBe: Alright :) Don't know how common of a case it'd be but I'm thinking one of these could be empty and the extending group would have it.

atluq: Not so common I guess. But I think it makes more sense to override if extending group has it, which covers that case.

generatedocs/compile/testdata/base/data/actions/changesheet/description.md
atluq: Can't tell really. Do we have this information somewhere else?

Will keep it in the testing-set.

atluq:
> Shouldn't the docs in the original location (generatedocs/data/actions) be removed?

No, I just copied the current one as a test-set. If that is what you mean.

generatedocs/compile/testdata/base/data/actions/changesheet/description.md
SeBBBe: Well we have the limitations: https://github.com/qlik-oss/gopherciser/blob/master/docs/README.md#limitations Which makes more sense to present, so I don't know why we'd also need this full list. Seems it'd just make it more difficult to maintain. I'll leave it up to @DnlLrssn

DnlLrssn: it's old docs that's there since before limitations section. In the beginning we only supported few objects and added them to this list. Can I agree it should probably be rewritten with a link to the limitations by now.

atluq: Yes, it fits better in limitations., but do we really want links to readme in the in-memory docs?

generatedocs/pkg/doccompiler/checks.go
```diff
@@ -0,0 +1,104 @@
...
+var checks = []check{
+	checkAllActionsDocumented,
+	checkAllConfigFieldsDocumented,
+	checkAllActionsInGroup,
+}
+
+func checkAndWarn(data *docData) {
+	for _, finding := range checkAll(data) {
+		fmt.Printf("WARNING: %v\n", finding)
+
```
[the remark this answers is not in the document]
atluq: Fixed!

generatedocs/pkg/doccompiler/compile.go
```diff
@@ -196,11 +248,48 @@ func (baseData *Data) overload(newData *Data) {
...
+	dirs := []string{}
+	for _, f := range files {
+		if f.IsDir() {
+			dirs = append(dirs, f.Name())
+		}
+	}
+	return dirs
+}
+
+// func groupNames(groups []common.GroupsEntry) []string {
```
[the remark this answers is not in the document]
atluq: Fixed.

atluq:
The biggest issue here is that markdown generation of parameters can be recursive. Solved this by demanding a struct tag `rec` on fields with recursive types used as stopping condition. I also added a stopping condition at depth 20, on which the generation exits with an error.

Suggestions on more robust solutions are welcome 😄

DnlLrssn:
we could possibly test if parent and child is of same type, but still wouldn't work in the case of  X -> Y -> X , in that case we would need to have an algorithm to detect circularity which seems like overkill for this. If we use, yet another, struct tag, we should probably prepare for possible other things we need in future and use the value, something like `docgen:"rec", and we could we could add, if we ever have a need for it, something e.g. add `docgen:"exclude"` if we need an exported field which should not be documented for some reason. Just making the example up, but we point is better to use a value of a tag so we can add other stuff easily in the future.

atluq:
> we could possibly test if parent and child is of same type, but still wouldn't work in the case of X -> Y -> X , in that case we would need to have an algorithm to detect circularity which seems like overkill for this.

Yes, it is overkill, if we need to implement from scratch. Tried this at first, but it became very hacky, and didn't cover the X -> Y -> X case. 

> If we use, yet another, struct tag, we should probably prepare for possible other things we need in future and use the value, something like `docgen:"rec", and we could we could add, if we ever have a need for it, something e.g. add `docgen:"exclude"` if we need an exported field which should not be documented for some reason. Just making the example up, but we point is better to use a value of a tag so we can add other stuff easily in the future.

True. Changing that.

DnlLrssn (approved):
very nice.
just needs to be documented in generatedocs/README.md

SeBBBe (approved):

PR #122, opened 2020-10-14
Enhancement/606 container object

DnlLrssn:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

- Add support for container objects
- Add new action `containertab` to switch object displayed in container
- Add support for new appstructure struct tag options
- Change some actions to use `selectable` list only in appstructure tag

**Info**


atluq (approved):
Looks really good 👍 

The action interface might be even clearer if we match mode value and config field. A suggestion is that the mode `id` and config field `activeid` could both be named `tabid`.

docs/settingup.md
```diff
@@ -478,6 +478,60 @@ A `ClickActionButton`-action simulates clicking an _action-button_. An _action-b
...
+
+### Settings
+
+* `mode`: Mode for container tab switching, one of: `id`, `random` or `index`.
+    * `id`: Switch to tab with object defined by `activeid`.
+    * `random`: Switch to a random tab within the container.
+    * `index`: Switch to tab with zero based index defined but `index`.
+* `id`: ID of the container object.
+* `activeid`: ID of the object to set as active, used with mode `id`.
+* `index`: Zero based index of tab to switch to, used with mode `index`.
```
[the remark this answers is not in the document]
DnlLrssn: makes sense to name the the same thing. tabid might be confusing though since it's the id of the object in the tab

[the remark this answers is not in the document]
DnlLrssn: and objectid might be confusing on top level with "id" hmm..

scenario/containertab.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	default:
+		return errors.Errorf("unknown container tab mode<%v>", settings.Mode)
+	}
+
+	return nil
+}
+
+// Execute ContainerTabSettings action (Implements ActionSettings interface)
+func (settings ContainerTabSettings) Execute(sessionState *session.State, actionState *action.State, connection *connection.ConnectionSettings, label string, reset func()) {
```
SeBBBe, line 97: No AffectAppObject method? o:

SeBBBe (approved):

scenario/containertab.go
DnlLrssn: Could add it to remove non active object in subsequent actions, BUT, 
1. All objects will still show for a select directly after changesheet (where container would have default object active)
2. Linked and masterobjects would be removed from subsequent actions even if they are included elsewhere on the sheet, in a different container where they are active, or subscribed to independetly.
So opted for not doing that

SeBBBe: Right 🤔

scenario/containertab.go
```diff
@@ -115,20 +128,46 @@ func (settings ContainerTabSettings) Execute(sessionState *session.State, action
 		activeID = settings.ObjectID
 	case ContainerTabModeRandom:
 		idx := sessionState.Randomizer().Rand(len(containerInstance.Children))
-		activeID = containerInstance.Children[idx].ObjID
+		child := containerInstance.Children[idx]
+		if child.External {
+			sessionState.LogEntry.Log(logger.WarningLevel, "")
```
SeBBBe, line 133: Empty log line? Also, is this not done as well in the isExternal method?

DnlLrssn: ... goland not working very good in wsl and didn't save it altough I pressed save... will update

DnlLrssn: fixed

SeBBBe (approved):

PR #134, opened 2020-10-28
Regression log

atluq:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Descriptive text which shall be used as squash commit message to master.

**Info**

The regression logging is done in the same logfile as the other logging. The regression log is implemented as a traffic log post-hook. The plan is to apply some filters on the traffic to limit I/O to the minimum of what is needed.


SeBBBe:
I think we should place some more sophisticated logic in the gopherciser to extract the data from objects since we have that in memory already, rather than logging everything. Otherwise it becomes an exercise in parsing JSON and applying logic to that on the qsest-diff side. The only thing we really care about is the data in the objects.

DnlLrssn:
Agree we should have some rules what to extract and only put relevant data into a regression log. As seems to be the idea with "reducedmessage" function. Since there's that thought of also adding possibility to diff meta data etc with the same set of tools, we might want to have log to a separate file with it's own predefined columns. A format which can be re-used by some form of meta-data extractor and possible other data interesting for regression analysis.

atluq:
> I think we should place some more sophisticated logic in the gopherciser to extract the data from objects since we have that in memory already, rather than logging everything. Otherwise it becomes an exercise in parsing JSON and applying logic to that on the qsest-diff side. The only thing we really care about is the data in the objects.

I will have another round at this then.

We dont have all the data we need in appstructure, right? Maybe we should log or store some extra data when calling `AffectsAppObjectsAction` 🤔

DnlLrssn:
> We dont have all the data we need in appstructure, right? Maybe we should log or store some extra data when calling `AffectsAppObjectsAction` 🤔

Regression should work without having an appstructure. And no, the appstructure won't have it in it's core data, which means that interface won't do much. Also wouldn't work when having e.g. random changesheet etc. So not the right place. Possibly object defintion with paths how to extract, OR just have a global set of extract paths, if we hit somehting on any and all of them , save it

SeBBBe:
Hmm yeah, the app structure doesn't really do the same thing. App evaluation, on the other hand.. 🤔  Although that action is in the app-evaluator code itself. That one walks through all objects and populates all the data. Then we'd want to filter it down and extract the data we care about the most. If we did something like that we wouldn't have to sort through websocket packets, match them together using IDs etc in order to find the relevant data as well as what object it belongs to.

PR #138, opened 2020-10-28
Add possibility to override script values

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

Add possibility to override script values

**Info**

https://github.com/qlik-oss/gopherciser/blob/enhancement/536-ScriptOverrides/docs/README.md#using-script-overrides

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -36,7 +36,7 @@ func AddConfigParameter(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&cfgFile, "config", "c", "", `Scenario config file.`)
 
 	// Script overrides
-	cmd.Flags().StringArrayVarP(&scriptOverrides, "set", "s", nil, "Override a value in script with key.path=value.")
+	cmd.Flags().StringArrayVarP(&scriptOverrides, "set", "s", nil, "Override a value in script with 'path/to/key=value'.")
 }
 
 // AddLoggingParameters add logging parameters to command
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -162,18 +165,64 @@ Sub-commands:
...
 
 `template` command flags:
 
 * `-c`, `--config string`: (optional) Create the specified scenario setup file. Defaults to `template.json`.
 * `-f`, `--force`: Overwrite existing scenario setup file.
 * `-h`, `--help`: Show the help for the `template` command.
 
+#### Using script overrides
+
+Script overrides overrides a value pointed to by a path to its key. If the key doesn't exist in the script there will be a not found error, even if it's a valid value according to config.
```
SeBBBe, line 184: A script override overrides a value..
or
Script overrides override values..

SeBBBe: not-found-error? "not found"-error?

cmd/helpers.go
```diff
@@ -52,14 +58,50 @@ func unmarshalConfigFile() (*config.Config, error) {
...
 
+func overrideScriptValues(cfgJSON []byte) ([]byte, []string, error) {
+	overrides := make([]string, 0, len(scriptOverrides))
+	for _, kvp := range scriptOverrides {
+		kvSplit := strings.SplitN(kvp, "=", 2)
+		if len(kvSplit) != 2 {
+			return cfgJSON, overrides, errors.Errorf("malformed override: %s, should be in the form key.path=value", kvp)
+		}
+		path := helpers.DataPath(kvSplit[0])
+		rawOrig, err := path.Lookup(cfgJSON)
```
SeBBBe, line 87: Still not entirely sure about this one 🤔  Also if the override is a json object, the individual keys won't be checked. I do understand it's a good thing to throw errors whenever possible, though, so removing it outright is not the best option either.

DnlLrssn: Not an easy one. My thinking, right or not, is that mistakes will more likely be done when writing the path, if an entire object is used it's at least more likely to be copy-pasted. My vote is on keeping the check, but agree not optimal, don't have a better solution.

DnlLrssn:
Should I add an option to read from file? Was thinking the format
```
key/to/path=value
key/to/path2=value2
key/to/pathN=valueN
```
then read line by line and execute as overrides.

docs/README.md
DnlLrssn: > If the key doesn't exist in the script there will an error, even if it's a valid value according to config.
?

cmd/helpers.go
SeBBBe: Yeah I tend to agree if the choice is between keeping it and removing it outright.

SeBBBe:
That sounds like a good idea. I can foresee a situation where there are too many overrides in the scheduler to be able to pass via the command..

SeBBBe:
Either that or the ability to pass overrides over stdin

DnlLrssn:
> 
> 
> Either that or the ability to pass overrides over stdin

Hmm true, will keep that in mind too

atluq:
> Should I add an option to read from file? Was thinking the format
> 
> ```
> key/to/path=value
> key/to/path2=value2
> key/to/pathN=valueN
> ```
> 
> then read line by line and execute as overrides.

It would also make sense for the file override to be ordinary json.

cmd/helpers.go
```diff
@@ -52,14 +58,50 @@ func unmarshalConfigFile() (*config.Config, error) {
...
+
 	return &cfg, nil
 }
 
+func overrideScriptValues(cfgJSON []byte) ([]byte, []string, error) {
+	overrides := make([]string, 0, len(scriptOverrides))
+	for _, kvp := range scriptOverrides {
+		kvSplit := strings.SplitN(kvp, "=", 2)
+		if len(kvSplit) != 2 {
+			return cfgJSON, overrides, errors.Errorf("malformed override: %s, should be in the form key.path=value", kvp)
```
markuswall, line 84: Here's another key.path reference

DnlLrssn: fixed

atluq:
There is actually a simple way of [checking non-existing fields when unmarshalling json](https://play.golang.org/p/MBlwJrAgsW4). This is not 100% related to this PR, but if we add this when unmarshalling the config in general we get some more safety when doing overrides.

DnlLrssn:
would not play nice with at least my way of work. I usually "comment" settings in actions by prefixing key with "_" or similar when I want temporarly want to use something else. Not sure how common that is though

DnlLrssn:
Also doesn't play well with the UnmarshalJSON interface which we use a lot of
https://github.com/golang/go/issues/41144

atluq:
> would not play nice with at least my way of work. I usually "comment" settings in actions by prefixing key with "_" or similar when I want temporarly want to use something else. Not sure how common that is though

Yes, this is a con.

> Also doesn't play well with the UnmarshalJSON interface which we use a lot of
> golang/go#41144

Yes, looks like the the creation of decoder has to be scattered all over the place. There must be a way not to 🤔

DnlLrssn:
will keep current implementation of checking the path as a form of error handling, since other kinds seems to get messy. This can be re-visited at a later point when we see some adoption and usage and better can determine issues.

SeBBBe (approved):
Looking good to me!

atluq (approved):
Looks good!

PR #140, opened 2020-11-03
Regression logging

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Add logs to be used for regression testing

Add configurable functionality which logs subscribed objects to a regression log file.

**Info**

Filters may need fine tuning.


scenario/actionhandler.go
```diff
@@ -459,9 +461,70 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
...
+			hyperCube := obj.HyperCube()
+			listObject := obj.ListObject()
+
+			err := sessionState.LogEntry.LogRegression(
+				// use unique id SESSION.ACTION.OBJ
+				fmt.Sprintf("%d.%d.%s", sessionState.LogEntry.Session.Session, sessionState.Counters.ActionID.Current(), obj.ID),
+				map[string]interface{}{
+					// "hyperCubeDataPages":  obj.HyperCubeDataPages(),
+					// "hyperCubeStackPages": obj.HyperCubeStackPages(),
+					// "hyperPivotPages":     obj.HyperPivotPages(),
```
DnlLrssn, line 512: adding these later or why commented? for many objects this is where the data will actually be

scenario/actionhandler.go
```diff
@@ -459,9 +461,70 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
...
+			if genObj, ok := obj.EnigmaObject.(*enigma.GenericObject); ok {
+				genType = genObj.GenericType
+			}
+
+			hyperCube := obj.HyperCube()
+			listObject := obj.ListObject()
+
+			err := sessionState.LogEntry.LogRegression(
+				// use unique id SESSION.ACTION.OBJ
+				fmt.Sprintf("%d.%d.%s", sessionState.LogEntry.Session.Session, sessionState.Counters.ActionID.Current(), obj.ID),
```
DnlLrssn, line 508: might want to get the actionid from log entry too, for the sake of container actions, other wise the end of the container action will have the actionID of the last subaction.

DnlLrssn (commented):
Might want to log in tsv format after all so it can easily used in SENSE
first X rows can be skipped in load script so filter row could stay, then headers then rows in tsv:
FILTERS ["- qNum","+ qStateCounts","+ qGrandTotalRow","+ qPivotDataPages","+ qStackedDataPages","+ qNum"]
ID    DATA    META
X      {...}       {...}

scenario/actionhandler.go
```diff
@@ -459,9 +461,70 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
 		containerActionEntry = originalActionEntry
 	}
 
-	err := logResult(sessionState, actionState, actionState.Details, containerActionEntry)
+	if err := logResult(sessionState, actionState, actionState.Details, containerActionEntry); err != nil {
+		return err
```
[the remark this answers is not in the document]
atluq: great catch!

scenario/actionhandler.go
atluq: thought there could be some redundancy in the data and left this for discussion.

scenario/actionhandler.go
atluq: I guess we should go with only the logEntry actionID then?

scenario/actionhandler.go
DnlLrssn: there will be for some objects yes, for others the data will not be there.

atluq:
> Might want to log in tsv format after all so it can easily used in SENSE
> first X rows can be skipped in load script so filter row could stay, then headers then rows in tsv:
> FILTERS ["- qNum","+ qStateCounts","+ qGrandTotalRow","+ qPivotDataPages","+ qStackedDataPages","+ qNum"]
> ID DATA META
> X {...} {...}

Yes, the existence of filter and readability made me go for no columns. But usability in Sense is probably more important.

scenario/actionhandler.go
DnlLrssn: I think that's safer bet.

DnlLrssn:
> 
> 
> > Might want to log in tsv format after all so it can easily used in SENSE
> > first X rows can be skipped in load script so filter row could stay, then headers then rows in tsv:
> > FILTERS ["- qNum","+ qStateCounts","+ qGrandTotalRow","+ qPivotDataPages","+ qStackedDataPages","+ qNum"]
> > ID DATA META
> > X {...} {...}
> 
> Yes, the existence of filter and readability made me go for no columns. But usability in Sense is probably more important.

yes readibiliy suffers somewhat, but makes it easier with sense. Enables excel/libreoffice and easier to use *nix commands like `cut`.

scenario/actionhandler.go
atluq: Uncommented all of them

scenario/actionhandler.go
```diff
@@ -461,16 +461,17 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
...
-		return err
-	}
+	logResultErr := logResult(sessionState, actionState, actionState.Details, containerActionEntry)
 	sessionState.LogEntry.LogDebugf("%s END", act.Type)
-
-	if err := logObjectRegressionData(sessionState); err != nil {
-		return err
+	logRegressionErr := logObjectRegressionData(sessionState)
+	var errs *multierror.Error
+	if logResultErr != nil {
```
DnlLrssn, line 468: if `errs` is of type multi.Error, which it is here, you shouldn't need to check the sub errors for nil, when appending
from multierror.Append:
````golang
if e != nil {
	err.Errors = append(err.Errors, e)
}
```

scenario/actionhandler.go
```diff
@@ -459,9 +461,67 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
...
 	sessionState.LogEntry.LogDebugf("%s END", act.Type)
-	return errors.WithStack(err)
+	logRegressionErr := logObjectRegressionData(sessionState)
+	var errs *multierror.Error
+	if logResultErr != nil {
+		errs = multierror.Append(errs, logResultErr)
+	}
+	if logRegressionErr != nil {
+		errs = multierror.Append(errs, logRegressionErr)
+	}
```
DnlLrssn, line 473:
```suggestion
		var errs *multierror.Error
		multierror.Append(errs, logResult(sessionState, actionState, actionState.Details, containerActionEntry))
		multierror.Append(logObjectRegressionData(sessionState))
```

DnlLrssn:
suggestion marking multiple lines didn't work very well when there was a comment in the middle of the lines I marked. So logResult is double now

atluq:
now that TestPanicAndFailRecover test fails again and I am unsure why, see `c69bfa1` build.

enigmahandlers/objectsmap.go
```diff
@@ -90,6 +90,25 @@ func (o *ObjectsMap) AddObject(obj *Object) error {
...
+	// 	fmt.Fprintf(os.Stderr, "genobj layout:\n%s\n", string(rawLayout))
+
+	// 	rawProps, err := typedObj.GetPropertiesRaw(context.TODO())
+	// 	if err != nil {
+	// 		panic(err)
+	// 	}
+	// 	fmt.Fprintf(os.Stderr, "genobj props:\n%s\n", string(rawProps))
+	// }
+	// fmt.Fprintln(os.Stderr, "-----------------------------")
+
```
DnlLrssn, line 111: removed commented code

logger/regression.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+)
+
+var filters = marshalFilters(
+	"+qStateCounts",
+	"+qGrandTotalRow",
+	"+qDataPages",
+	"+qPivotDataPages",
+	"+qStackedDataPages",
+	"-qNum",
```
DnlLrssn, line 43: So this these seems to be filters for _hypercube_ top level, so at least "qListObject" for _listObject_ top level should be added.  Should the rest "top level" data also be in filter?
"hyperCubeDataPages"
"hyperCubeStackPages"
"hyperCubePivotPages"
"listObjectDataPages"
as all data in these should be considered significant.

scenario/actionhandler.go
```diff
@@ -459,8 +461,63 @@ func (act *Action) endAction(sessionState *session.State, actionState *action.St
...
+					"hyperCube":           obj.HyperCube(),
+					"listObject":          obj.ListObject(),
+					"listObjectDataPages": obj.ListObjectDataPages(),
+				},
+				map[string]interface{}{
+					"actionType":  sessionState.LogEntry.Action.Action,
+					"actionLabel": sessionState.LogEntry.Action.Label,
+					"actionID":    sessionState.LogEntry.Action.ActionID,
+					"objectID":    obj.ID,
+					"objectType":  genObj.Type,
```
DnlLrssn, line 515: genObj.Type will almost always be GenericObject, genObj.GenericType might be more useful.

logger/regression.go
SeBBBe: The filters above match what we have for the SDKExerciser regression logging. Not saying that means it's perfectly correct. We might have to play around with the filters once we have the actual differ to see that they have the intended results.

logger/regression.go
```diff
@@ -0,0 +1,93 @@
...
+// NewRegressionLogger creates a new RegressionLoggerCloser with headerEntries
+// written in the header of the log.
+func NewRegressionLogger(w io.WriteCloser, headerEntries ...HeaderEntry) RegressionLoggerCloser {
+	fmt.Fprintf(w, "HEADER_KEY\tHEADER_VALUE\n")
+	fmt.Fprintf(w, "FILTERS\t%s\n", filters)
+	for _, he := range headerEntries {
+		fmt.Fprintf(w, "%s\t%s\n", strings.ToUpper(strings.TrimSpace(he.Key)), strings.TrimSpace(he.Value))
+	}
+	fmt.Fprintln(w, "---")
+	fmt.Fprintln(w, "ID\tMETA\tDATA")
```
SeBBBe, line 63: So we'll have a tsv with different amount of columns depending on row position? Is this format digestible by Sense etc?

SeBBBe (commented):
Looks pretty good!

scenario/actionhandler.go
atluq: Yes, agree. Tried both before and dont know why this one ended up in the metadata.

logger/regression.go
atluq: Well, if we could start reading from after `---`. I have not tried yet.

logger/regression.go
DnlLrssn: The data the filter in SDKExerciser is working on looks different, and my comments wouldn't apply there, except for the qListObject one.

atluq: If we know data under these top level keys shall be fully included, I can add this now. 

Then we can fine tune as we have the diff tool.

logger/regression.go
DnlLrssn: you can set the load script in sense to skip the first X rows

logger/regression.go
DnlLrssn: for many objects the actual data will be in e.g. _hyperCubeDataPages_ not _hypercube_ using this structure, this is different from how it works in SDKExerciser which works directly on traffic

DnlLrssn: only objects which gets their data in _GetLayout_ request will have the significant portion of data in _hypercube_. This is a large chunk of the objects though.

logger/regression.go
atluq: Okay. Then we would have to fix the number of rows in the first part, or chose number of rows dynamically searching for "---". 

My idea was for us to be able to add new metadata rows in the first part (like `ID_FORMAT`).

logger/regression.go
atluq: Okay, we still need both 👍 

And `hyperCubeDataPages/**` shall all be included?

logger/regression.go
DnlLrssn: yea fixed would be better from sense perspective

logger/regression.go
atluq: Added some unit tests where we can test the intended behavior for real logs.

DnlLrssn (approved):
Other than the filters which we can update at a later point

DnlLrssn:
@atluq please hold of merging this until next release has been done, there might be more bugfixes incoming before release is set.

atluq:
> @atluq please hold of merging this until next release has been done, there might be more bugfixes incoming before release is set.

Yes. Will not merge until the full pipe of regression analysis is ready. It unnecessary to release this until then.

SeBBBe (approved):

SeBBBe:
Selecting schedulers is not fully supported in the GUI, just FYI. We can probably get it working but it will require some manual effort.

atluq:
> Selecting schedulers is not fully supported in the GUI, just FYI. We can probably get it working but it will require some manual effort.

Great to know. This commit disables forced scheduler and automatically overrides any scheduler with a preconfigured simpleScheduler, both only if regression logging is enabled. So no manual scheduler selection is needed.

SeBBBe:
> > Selecting schedulers is not fully supported in the GUI, just FYI. We can probably get it working but it will require some manual effort.
> 
> Great to know. This commit disables forced scheduler and automatically overrides any scheduler with a preconfigured simpleScheduler, both only if regression logging is enabled. So no manual scheduler selection is needed.

Ah, right. Didn't look at it carefully enough I guess. That's a good way of doing it - we should add logic in the frontend to hide the irrelevant settings when the checkbox is ticked.

atluq:
> we should add logic in the frontend to hide the irrelevant settings when the checkbox is ticked.

Good point, will think of a good way to do that 👍

scheduler/regression.go
```diff
@@ -0,0 +1,21 @@
...
+			Scheduler: Scheduler{
+				SchedType:      SchedSimple,
+				InstanceNumber: 0,
+			},
+			Settings: SimpleSchedSettings{
+				ExecutionTime:   -1,
+				Iterations:      1,
+				ConcurrentUsers: 1,
+				RampupDelay:     0.5,
+			},
```
DnlLrssn, line 17: init SimpleScheduler as 

```golang
&SimpleScheduler{Scheduler{SchedSimple,0,}, SimpleSchedSettings{.etc}
```

And we would get a compile error if we add an entry = not forget to add that here.

SeBBBe (approved):
👍

scheduler/regression.go
atluq: Changing them all reads pretty bad 🤔
```
			Scheduler{
				SchedSimple,
				TimeBuffer{},
				0,
				session.ReconnectSettings{},
				nil,
				false,
			},
			SimpleSchedSettings{
				-1,
				1,
				0.5,
				1,
				false,
				false,
			},
```

PR #141, opened 2020-11-06
Support containers with show conditions

DnlLrssn:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

* Support containers with show conditions
* Automatically switch child if current no longer considered visible
* If no visible child remove current subscription
* Update `containertab` random to only pick among visible children
* error on userdefined containertab switch to non-visible children

**Info**

### DIFF
--- a/session/containerhandler.go
+++ b/session/containerhandler.go
@@ -29,7 +29,7 @@ type (
 	ContainerHandlerInstance struct {
 		ID       string
 		ActiveID string
-		Children []ContainerChildReference
+		children []ContainerChildReference
 
 		lock sync.Mutex
 	}
@@ -97,6 +97,7 @@ func (handler *ContainerHandlerInstance) SetObjectAndEvents(sessionState *State,
 
 	// Get layout on object changed
 	event := func(ctx context.Context, as *action.State) error {
+		sessionState.LogEntry.Logf(logger.DebugLevel, "Getting layout for object<%s> handle<%d> type<%s>", genObj.GenericId, genObj.Handle, genObj.GenericType)
 		layout := GetContainerLayout(sessionState, as, genObj)
 		if as.Failed {
 			return nil // error occured, but has been reported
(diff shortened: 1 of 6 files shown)


session/containerhandler.go
```diff
@@ -39,6 +46,8 @@ type (
 		IsMaster          bool               `json:"isMaster"`
 		ExternalReference *ContainerExternal `json:"externalReference"`
 		Type              string             `json:"type"`
+		// Condition fulfilled  show condition. Value is nil on no show condition, true if it should show and false otherwise
+		Condition *string `json:"condition"`
```
atluq, line 50: this physically hurts🩹

atluq (commented):
Looks good! I guess docs need an update, so will have another look then.

What is the reasoning behind picking from visible only for the random mode of the action?

session/containerhandler.go
DnlLrssn: > 
> 
> Looks good! I guess docs need an update, so will have another look then.
> 
> What is the reasoning behind picking from visible only for the random mode of the action?

user can only see the visible, there's no tab showing for visualizations not visible, so reason is to chose only among those tabs a user would use.

Don't think docs need an update, maybe some clarification?

DnlLrssn: you mean the tristate, or the fact they've chosen to send a string?

DnlLrssn: should probably make the comment "True" and "False", will update that

atluq: Maybe throw in a "visible" [here](https://github.com/qlik-oss/gopherciser/blob/80bb4fc10a278d1f8b85431959c11c17ffb1bb60/generatedocs/data/params.json#L225).

I meant why we dont chose from visible in all modes. But it makes sense to leave it up to the script author, when the mode is not random.

DnlLrssn: sure can clarify that part. But as a user I would assume "random" only select a selectable tab.

atluq: Meant that it is a string in the message.

atluq (approved):

session/containerhandler.go
DnlLrssn: > 
> 
> Maybe throw in a "visible" [here](https://github.com/qlik-oss/gopherciser/blob/80bb4fc10a278d1f8b85431959c11c17ffb1bb60/generatedocs/data/params.json#L225).
> 
done. now I can check the doc checkbox too ;)

atluq: 😆

PR #144, opened 2020-11-11
Enhancement/701 support more session objects

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Create 3 more session objects on open app for api compliance
- StoryList
- VariableList
- LoadModelList

**Info**

### DIFF
--- a/senseobjects/app.go
+++ b/senseobjects/app.go
@@ -24,6 +24,7 @@ type (
 		localeInfo        *enigma.LocaleInfo
 		variablelist      *VariableList
 		storylist         *StoryList
+		loadmodellist     *LoadModelList
 		mutex             sync.Mutex
 	}
 )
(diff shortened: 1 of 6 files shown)


senseobjects/app.go
```diff
@@ -123,6 +131,145 @@ func (app *App) setBookmarkList(sessionState SessionState, bl *BookmarkList) {
 	app.bookmarkList = bl
 }
 
+// GetVariableList create or return existing variable list session object
+func (app *App) GetVariableList(sessionState SessionState, actionState *action.State) (*VariableList, error) {
```
SeBBBe, line 135: This function could benefit from being made generic with the help of an interface and/or a function argument.

SeBBBe (approved):
It feels a little bit like reading the same code change 3 times 😛

senseobjects/app.go
DnlLrssn: was thinking about it, but for most cases we need / or will need when accessing details, specific structs for the layout and properties(like e.g. sheetlist and bookmarklist), so would be little gain in the end

DnlLrssn: or you mean GetSessionObject(sdslfsdf) and we autodetect the rest? semiprepared for that by separating the current session objects to its own struct. Would agree a "sessionobjecthandler" or something shoule be implemented at some point.

DnlLrssn:
> 
> 
> It feels a little bit like reading the same code change 3 times 😛

values, input and output types different though, kind of missing the c# "abstract" class ...

atluq (commented):
Agree with Fabian that an interface is appropriate for VariableList, StoryList etc.

atluq:
> > It feels a little bit like reading the same code change 3 times 😛
> 
> values, input and output types different though, kind of missing the c# "abstract" class ...

Could use composition for this. VariableList, StoryList and LoadModelList could share the same core struct.

senseobjects/app.go
DnlLrssn: would only work for these three and only while they use the GenericLayout and GenericProperties, as soon as we have to make any of them their own type the need to be broken out again, since this has been the case with the other session objects I consider it a likely scenerio to happen. Also remember

layout := MyLayout{}
myfunc(&layout)

func myfunc(layout interface{}{ 
...
json.Unmarshal(data, layout)
...
}
will make layout into a map[string] instead of a MyLayout

DnlLrssn: that said, some code could be shared, but would rather make something that covers "all" sessionobjects in that case, which would be a bigger rework then I wanted to do now.

atluq (approved):
Duplication aside 😄 it all looks good to me.  Lets look at the duplication in another PR.

DnlLrssn:
> I'm thinking of something like:
>
>type interface SenseObj {
>    UpdateLayout(context.Context) error
>    UpdateProperties(context.Context) error
>}

sure, but something like that should be used as a bigger re-factor for all the session objects.

DnlLrssn:
> Duplication aside 😄

C# developer in me as screaming, but yea after all, absence of  generics means use one of:
1. interfaces
2. type assertions
3. copy paste
option 3 is almost to be prefered if there's few implementations needed, but now we're starting to get quite some session objects so should be refactored

atluq:
> > Duplication aside 😄
> 
> C# developer in me is screaming, but yea after all, absence of generics means use one of:
> 
> 1. interfaces
> 2. type assertions
> 3. copy paste
>    option 3 is almost to be prefered if there's few implementations needed, but now we're starting to get quite some session objects so should be refactored

Agree, copy paste is definitely the best solution at many times .

Composition can also remove some duplication, but often feels a bit forced.

DnlLrssn:
yes true, but then you get the types of structs enigma has, not sure if I like that

PR #147, opened 2020-11-20
Cover race condition sometimes triggering in very large simulations.

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Remove async sense websocket disconnect listener
* Add sync execute "fail" functions
* make ws fail fuction async
* move pending handler to its own package
* double default backoff times

**Info**

This makes some parts of the the code more synchronous to cover a race condition which can happen when running larger simulations of 20 - 40 000 simulated users or so. 

Moved the pending handler to its own package in a different solution. No longer needed with current implementation, but kept the package since probably belong in its own package better.


pendinghandler/pendinghandler.go
```diff
@@ -1,4 +1,4 @@
-package session
+package pending
```
SeBBBe, line 1: package name does not correspond to folder name

SeBBBe (approved):

pendinghandler/pendinghandler.go
DnlLrssn: yea I know, didn't like naming the folder just `pending` so opted to break convention and name it `pendinghandler`. didn't want usage to be pendinghandler.Handler or pendinghandler.PendingHandler so felt package name should be `pending`. I know it's breaking convention, but I take it your opinion is to stick to convention and name the folder `pending` then?

SeBBBe: That's prefereable imo. It could also be renamed, for instance to something like `pendingrequests` so then it'd be `pendingrequests.Handler`

atluq (approved):
Looks great!

foh79 (approved):
lgtm

PR #152, opened 2020-12-02
Bugfix/710 artifact map

DnlLrssn:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

Refactor artifact map to be able to better support new resource types and avoid incorrectly using non app resources in app actions.

**Info**

some unknowns:
* Should we clear entire artifact map in elasticexplore,current assumption is yes

### DIFF
--- a/session/artifactmap.go
+++ b/session/artifactmap.go
@@ -27,11 +27,14 @@ type (
 	}
 
 	// ArtifactList all items, spaces and collections
-	ArtifactList []*ArtifactEntry
+	ArtifactList struct {
+		list   []*ArtifactEntry
+		sorted bool
+	}
 
 	// ArtifactMap is the map between app names and GUIDs
 	ArtifactMap struct {
-		resourceMap map[string]ArtifactList
+		resourceMap map[string]*ArtifactList
 		mu          sync.RWMutex
 	}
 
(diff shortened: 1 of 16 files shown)


session/artifactmap.go
```diff
@@ -12,175 +12,191 @@ import (
...
-		streamTitleToID:  &sync.Map{},
-		appTitleToItemID: &sync.Map{},
-		spaceTitleToID:   &sync.Map{},
-	}
+	return &ArtifactMap{resourceMap: make(map[string]*ArtifactList)}
 }
 
-// IsEmpty returns true of entry is nil or has no GUID
-func (entry *ArtifactEntry) IsEmpty() bool {
+// Copy of ArtifactEntry
```
atluq, line 94: Data is not copied. Add this info to doc-comment if this is on purpose.

session/artifactmap.go
```diff
@@ -12,175 +12,191 @@ import (
...
 }
-func (d ArtifactListDict) Swap(i, j int) {
-	d[i], d[j] = d[j], d[i]
+
+func (d ArtifactList) Swap(i, j int) {
+	d.list[i], d.list[j] = d.list[j], d.list[i]
 }
 
 // NewAppMap returns an empty ArtifactMap
 func NewAppMap() *ArtifactMap {
```
atluq, line 90: Refactor to NewArtifactMap?

atluq (commented):
Seems to fix the issue and improves code quality in general 👍

session/artifactmap.go
DnlLrssn: no, just missed than one when adding .Data, will fix

session/artifactmap.go
DnlLrssn: yepp missed that one,

SeBBBe (approved):

session/artifactmap.go
```diff
@@ -12,175 +12,192 @@ import (
...
 	}
 
-	// todo how to handle itemID?
+	return &ArtifactEntry{
+		Name:         entry.Name,
+		ID:           entry.ID,
+		ItemID:       entry.ItemID,
+		ResourceType: entry.ResourceType,
+		Data:         entry.Data,
+	}
```
atluq, line 106: This would remove the need of remembering to update this function:
```
entryShallowCopy := *entry
return &entryShallowCopy

atluq (approved):

session/artifactmap.go
atluq: It would also be a horrible mistake in C 😄

DnlLrssn: will work as long as we don't add a pointer, slice, array or map.

DnlLrssn: but in current version yes would work, made it this way out of habit, since we usually have something which has a pointer, but probably better in this case

atluq: true, the way it is now, a bug like that would be easier to find

either way for me 👍

DnlLrssn: changed to the cleaner looking version

foh79 (approved):
Tested various openhub/explore with links, datasets and apps + compared loadclient CPU/RAM

PR #160, opened 2020-12-16
Enhancement/709 sheet show condition

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Return error on `changesheet` action if the sheet is considered hidden
* Don't randomize changesheet to hidden sheets when using random action.

**Info**

### DIFF
--- a/helpers/stringbool.go
+++ b/helpers/stringbool.go
@@ -1,0 +1,28 @@
+package helpers
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type (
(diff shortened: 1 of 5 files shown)


SeBBBe (approved):

atluq (approved):
LGTM

Long live the StringBool!

Mounikabanka (approved):
Works fine on the build show sheet condition turned on.

Note:

- GUI shows all the hidden sheets as well.
- Any old build such as (April 2020) can view all the hidden sheets as well i.e., the client is old to understand the flag and Tool throws the error in the old build as well because the structure is saved to the app.

PR #165, opened 2021-01-11
Change test connection towards elastic to a new URL.

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Change test connection towards elastic to a new URL.

**Info**


atluq (approved):
Should be ok as long as there at this point is no concurrent use of the rest client 👍

DnlLrssn:
> 
> 
> Should be ok as long as there at this point is no concurrent use of the rest client 👍

yea there shouldn't be during test connection, the connection functions are sequential.

SeBBBe (approved):
Nice!

PR #174, opened 2021-01-27
add /debug/pprof endpoint when running tool with --metrics flag

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add /debug/pprof endpoint to tool when running a http listener (currrently done with the --metrics flag) for debugging purposes. Should not affect anything other then slightly bigger binary unless listener is running and endpoint accessed.

**Info**

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	_ "net/http/pprof"
+
 	"github.com/qlik-oss/gopherciser/cmd"
 )
 


SeBBBe (approved):
It just terminates if there is no web server?

DnlLrssn:
> 
> 
> It just terminates if there is no web server?

Has far as I could understand it should not do any profiling unless the endpoint is called, i.e. with no webserver shouldn't impact. The import has an init which registers some handlers to default https muxer. (unfortunatly no way to avoid having that init run). And to my understanding shouldn't be run. TBH couldn't conclusively  find this to be the case, hence the use of the word "should".

DnlLrssn:
calls are e.g. 
http://localhost:[metricsport]/debug/pprof/profile?seconds=30 to do 30s CPU profile

SeBBBe:
That's the impression I got as well from a quick googling. Just wanted to make sure that it won't kick off any profiling when running normally.

atluq (approved):
LGTM!

PR #175, opened 2021-01-27
Add handling of feature flags

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add feature flag map to session state to enable logic flows differing depending on feature flags enabled.

**Info**

### DIFF
--- a/session/features.go
+++ b/session/features.go
@@ -1,0 +1,64 @@
+package session
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/pkg/errors"
+	"github.com/qlik-oss/gopherciser/action"
+	"github.com/qlik-oss/gopherciser/logger"
+)
+
+type (
+	// Features maps feature flags and values
+	Features struct {
+		m  map[string]bool
+		mu sync.Mutex
+	}
+)
+
+// UpdateFeatureMap request features from server and updates feature map
+func (features *Features) UpdateFeatureMap(rest *RestHandler, host string, actionState *action.State, logEntry *logger.LogEntry) {
+	features.mu.Lock()
+	defer features.mu.Unlock()
+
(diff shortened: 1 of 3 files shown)


SeBBBe (approved):
I assume this is to be used in an upcoming PR?

DnlLrssn:
> 
> 
> I assume this is to be used in an upcoming PR?

To be used either in upcoming PR or derivates using gopherciser as library, tbd where.

SeBBBe:
Not a big fan of adding features with no planned use. Code that's not being used should generally be removed.

That said though, this is clean code and seems like it might come in handy.

DnlLrssn:
> 
> 
> Not a big fan of adding features with no planned use. Code that's not being used should generally be removed.
> 

There's a planned use, just not decided where.

SeBBBe:
Sounds like an oxymoron to me.

session/features.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if err := jsonit.Unmarshal(req.ResponseBody, &features.m); err != nil {
+		actionState.AddErrors(errors.WithStack(err))
+		return
+	}
+}
+
+// IsFeatureEnabled check if feature flag is enabled
+func (features *Features) IsFeatureEnabled(feature string) (bool, error) {
```
atluq, line 41: We could return two different error types here to be able to recover in different ways, if feature map is not fetched or feature does not exist.

Do not know if it will be needed though ^^

atluq (approved):
Clean code! Not too much to add to previous discussion ;)

session/features.go
DnlLrssn: yea probably won't be used, but easy to add so can do that

PR #180, opened 2021-02-11
Elasticreload api-compliance

atluq:
**Squash commit message**

Make elasticreload api compliant

Add http requests:
GET api/v1/items/ITEMID
GET /api/v1/reloads?appId=APPID

### DIFF
--- a/scenario/elasticreload.go
+++ b/scenario/elasticreload.go
@@ -26,6 +26,7 @@ type (
 const (
 	postReloadEndpoint = "api/v1/reloads"
 	getReloadEndpoint  = "api/v1/reloads"
+	getItemsEndpoint   = "api/v1/items"
 
 	// Delay time between re-connect of event websocket and checking status page if reload is still not done
 	StatusCheckDelay = 30 * time.Second
@@ -117,6 +118,15 @@ func (settings ElasticReloadSettings) execute(sessionState *session.State, actio
 			reloadEventChan <- &event
 		}
 	}, true)
+
+	itemsUpdatedFunc := events.RegisterFunc(eventws.OperationUpdated, func(event eventws.Event) {
+		if !helpers.IsContextTriggered(statusContext) && event.ResourceType == eventws.ResourceTypeItems {
+			_ = sessionState.Rest.GetAsync(
+				fmt.Sprintf("%s/%s/%s", host, getItemsEndpoint, event.ResourceID),
+				actionState, sessionState.LogEntry, nil)
+		}
+	}, true)
+
 	// Re-use event structure to "listen" on websocket re-connecting
 	wsReconnectFunc := events.RegisterFunc(session.EventWsReconnectEnded, func(event eventws.Event) {
 		// If event websocket was re-connected during reload, wait "StatusCheckDelay" then check status page if reload event still hasn't triggered to make sure reload is still ongoing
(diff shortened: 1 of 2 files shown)


SeBBBe:
Kind of depends what type of reload we want to test. I assume this is the kind where the user is standing on the "Reload history" tab? What about when doing it through the meatball menu? Or doing it on a scheduled basis? I assume those make up the bulk of reloads so maybe we should introduce a parameter to be able to choose between different types as the traffic generated will be quite different.

atluq:
> Kind of depends what type of reload we want to test. I assume this is the kind where the user is standing on the "Reload history" tab? What about when doing it through the meatball menu? Or doing it on a scheduled basis? I assume those make up the bulk of reloads so maybe we should introduce a parameter to be able to choose between different types as the traffic generated will be quite different.

That is a good idea.

Reloading from hub view and reload history view have this request flow. Have not looked closer at scheduled reloads.

SeBBBe:
Ah they both have it? I thought it would GET items unless the user clicks on the toast.

DnlLrssn:
> 
> 
>  Or doing it on a scheduled basis? 

Scheduled should not be tested through tool though, that's better done manually as the schedule part will take care of the multiple runs.

scenario/elasticreload.go
```diff
@@ -177,6 +188,11 @@ forLoop:
 					actionState.AddErrors(errors.Errorf("reload result event contains reload id of unexpected type<%T> value<%v>", eventReloadIDEntry, eventReloadIDEntry))
 					break forLoop
 				}
+
+				_ = sessionState.Rest.GetAsync(
+					fmt.Sprintf("%s/%s?appId=%s", host, getReloadEndpoint, event.ResourceID),
+					actionState, sessionState.LogEntry, nil)
```
DnlLrssn, line 194: Isn't this the part they really really didn't want us to do? it's the result on the reload right? which you only would see if your on the reload details page? The argument was that most people won't be.

DnlLrssn: same with the one added to event.

scenario/elasticreload.go
```diff
@@ -117,6 +118,15 @@ func (settings ElasticReloadSettings) execute(sessionState *session.State, actio
 			reloadEventChan <- &event
 		}
 	}, true)
+
+	itemsUpdatedFunc := events.RegisterFunc(eventws.OperationUpdated, func(event eventws.Event) {
+		if !helpers.IsContextTriggered(statusContext) && event.ResourceType == eventws.ResourceTypeItems {
+			_ = sessionState.Rest.GetAsync(
+				fmt.Sprintf("%s/%s/%s", host, getItemsEndpoint, event.ResourceID),
+				actionState, sessionState.LogEntry, nil)
```
DnlLrssn, line 126: same with this one

scenario/elasticreload.go
DnlLrssn: >Reloading from hub view and reload history view have this request flow. Have not looked closer at scheduled reloads.

is even this one in hub now even if user stays in hub until reload is done?

atluq: It is a list of reload results. It is done also when reloading from the meatball menu in the hub.

DnlLrssn (commented):
code looks good, but want to have it verified the items request on every event and at end also happens if users is on hub landing page (or in explore page).

scenario/elasticreload.go
DnlLrssn: > 
> 
> It is a list of reload results. It is done also when reloading from the meatball menu in the hub.

and where are not talking about the meatball menu in "details" right?

SeBBBe (approved):

PR #183, opened 2021-02-16
Remove logging of session ID.

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Remove session name from log entries.

**Info**

### DIFF
--- a/enigmahandlers/senseconnect.go
+++ b/enigmahandlers/senseconnect.go
@@ -164,13 +164,6 @@ func (uplink *SenseUplink) Connect(ctx context.Context, url string, headers http
 		case "ws":
 			u.Scheme = "http"
 		}
-
-		for _, v := range dialer.Jar.Cookies(u) {
-			// A "normal" connection has cookie name "X-Qlik-Session" a virtual proxy has configurable cookie name, as long as the name configured starts with "X-Qlik-Session" it will be logged
-			if v != nil && strings.HasPrefix(v.Name, "X-Qlik-Session") {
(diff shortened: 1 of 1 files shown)


atluq (commented):
Shall we remove the SessionName field from SessionEntry struct or give it a value from another source?

SeBBBe:
It will still log a column though? The AllFields that's passed in to CreateTSVLogger remains unchanged.

atluq:
> Shall we remove the SessionName field from SessionEntry struct or give it a value from another source?

Removing it would probably break a few pipelines, so I see why it is kept.

atluq (approved):

DnlLrssn:
> 
> 
> > Shall we remove the SessionName field from SessionEntry struct or give it a value from another source?
> 
> Removing it would probably break a few pipelines, so I see why it is kept.

It would break sense loadscripts defining this column, so that's the exact reason why it's still there.

SeBBBe:
Then why even make this change though? If the session id is salted it could still be used as a unique key, theoretically.

DnlLrssn:
> 
> 
> Then why even make this change though? If the session id is salted it could still be used as a unique key, theoretically.

yes, but we'd have to get the salt and hash it. And then we need to know if it's QSEoW to know if we can send requests to figure out the salt, which would be a lot more code changes has this part of the code has no clue what kind of environment we are connecting to, it could be QSEoK, QSEoW, Qlik Core etc etc. Would also need to start mixing in REST requests into the websocket connect logic which I'm not very keen on either.

PR #184, opened 2021-02-16
"stringbools" can sometimes be sent as "0" instead of "False"

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

string "boolean" from engine under certain conditions sent as "0" instead of "False".

**Info**


SeBBBe (approved):
Would be much better in javascript! ;)

Mounikabanka (approved):
looks good

PR #186, opened 2021-02-17
Add prometheus metrics for API calls

SeBBBe:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

Add a prometheus SummaryVec and Histogram for duration per HTTP-level API calls including path, method and status code. Limitation: only aggregates results to one level after "api/v1", e.g.

GET https://test:9001/api/v1/evaluations -> "api/v1/evaluations", GET, 200
POST https://test:9001/api/v1/evaluations/abc123 -> "api/v1/evaluations", POST, 201
GET https://test:9001/api/dcaas -> no reporting

### DIFF
--- a/metrics/custom.go
+++ b/metrics/custom.go
@@ -15,7 +15,7 @@ var ApiCallDuration = prometheus.NewHistogramVec(
 		Help:    "A histogram of HTTP request durations.",
 		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.25, 0.5, 1., 2.0, 2.5, 5., 10.},
 	},
-	[]string{"path", "method", "status_code"},
+	[]string{"action", "path", "method", "status_code"},
 )
 
 // ApiCallDurationQuantile summary for API call duration
@@ -25,7 +25,7 @@ var ApiCallDurationQuantile = prometheus.NewSummaryVec(
 		Help:       "A summary of HTTP request durations",
 		Objectives: map[float64]float64{0.1: 0.1, 0.5: 0.05, 0.95: 0.01, 0.99: 0.001, 0.999: 0.0001},
 	},
-	[]string{"path", "method", "status_code"},
+	[]string{"action", "path", "method", "status_code"},
 )
 
 // GopherActions action counter
--- a/metrics/custom.go
+++ b/metrics/custom.go
@@ -8,6 +8,26 @@ const (
 	promNS = "gopherciser"
 )
 
+// ApiCallDuration histogram for API call duration
+var ApiCallDuration = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Name:    "api_request_duration_seconds",
+		Help:    "A histogram of HTTP request durations.",
+		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.25, 0.5, 1., 2.0, 2.5, 5., 10.},
+	},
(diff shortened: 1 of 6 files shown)


metrics/custom.go
```diff
@@ -8,6 +8,26 @@ const (
...
+	},
+	[]string{"path", "method", "status_code"},
+)
+
+// ApiCallDurationQuantile summary for API call duration
+var ApiCallDurationQuantile = prometheus.NewSummaryVec(
+	prometheus.SummaryOpts{
+		Name:       "api_request_duration_quantiles_seconds",
+		Help:       "A summary of HTTP request durations",
+		Objectives: map[float64]float64{0.1: 0.1, 0.5: 0.05, 0.95: 0.01, 0.99: 0.001, 0.999: 0.0001},
```
SeBBBe, line 26: Is this correct? Copy-pasted from app-eval, although it looks a bit strange to me.

HvonP: Looks superweird, but seems consistent with this
https://github.com/prometheus/client_golang/blob/babeb356a51b1277535260e0122ad8c46240ab96/prometheus/examples_test.go#L311

HvonP (approved):

metrics/custom.go
```diff
@@ -8,6 +8,26 @@ const (
...
 )
 
+// ApiCallDuration histogram for API call duration
+var ApiCallDuration = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Name:    "api_request_duration_seconds",
+		Help:    "A histogram of HTTP request durations.",
+		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.25, 0.5, 1., 2.0, 2.5, 5., 10.},
+	},
+	[]string{"path", "method", "status_code"},
```
DnlLrssn, line 18: we should also have action as a label

SeBBBe:
Added action/label. But.. do we really want it? Not so 1337 with grafana, but is there a way to aggregate per endpoint while ignoring one of the labels? In case we have a ton of different labels in action for example, otherwise there could be a scenario where the data becomes too sparse.

DnlLrssn:
> 
> 
> Added action/label. But.. do we really want it? Not so 1337 with grafana, but is there a way to aggregate per endpoint while ignoring one of the labels? In case we have a ton of different labels in action for example, otherwise there could be a scenario where the data becomes too sparse.

Have no opinion if we want it, it's what was decided upon meeting, have no personal option though. @alfinhaji ?

DnlLrssn (approved):

alfinhaji:
> > Added action/label. But.. do we really want it? Not so 1337 with grafana, but is there a way to aggregate per endpoint while ignoring one of the labels? In case we have a ton of different labels in action for example, otherwise there could be a scenario where the data becomes too sparse.
> 
> Have no opinion if we want it, it's what was decided upon meeting, have no personal option though. @alfinhaji ?

Im not sure at this time, let see what the data shows and go from there.  I think there may be some use to having action label included

alfinhaji:
So looks like we are adding ~2.5K metrics per load cannon with API data which @jonathansm1th believes is fine.  Also, having the action/label I think is useful, it allows us to build a dashboard with will show something like this

<Service>
      <Action>
             <API>
i.e. we can split of the latency of the /api/v1/collections API in eslaticopenhub versus elasticexplore versus createapp versus createcollection....etc

Also validated metrics are reaching the push gateway and available in grafana
image
image
image

PR #187, opened 2021-02-17
Make elasticmoveapp navigate to space by default

atluq:
**Checklist**

- [ ] documentation updated

**Squash commit message**

Make elasticmoveapp navigate to space by default

### DIFF
--- a/docs/settingup.md
+++ b/docs/settingup.md
@@ -219,6 +219,7 @@ Apply a bookmark in the current app.
 
 * `title`: Name of the bookmark (supports the use of [variables](#session_variables)).
 * `id`: ID of the bookmark.
+* `selectionsonly`: *Missing documentation*
 
 ### Example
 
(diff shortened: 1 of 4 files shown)


DnlLrssn (commented):
If we navigate to a new space I would expect the artifactmap to be filled the the apps from the space we navigate to so apps there could be opened. Filling with new apps would have two more consequences, 
* we should add a flag to keep or not to keep current artifacts (like in elasticexplore)
* adding apps to artifactmap will break the existing flow of scripts,

atluq:
> If we navigate to a new space I would expect the artifactmap to be filled the the apps from the space we navigate to so apps there could be opened. Filling with new apps would have two more consequences,
> 
> * we should add a flag to keep or not to keep current artifacts (like in elasticexplore)
> * adding apps to artifactmap will break the existing flow of scripts,

Right! I see two ways to go:
    1. Move to space and add apps to artifactmap, but not by default (opposite of the default in browser-client).
    2. Do not merge this pr.

SeBBBe:
I think it should navigate by default so that default matches the normal user flow.

atluq:
> I think it should navigate by default so that default matches the normal user flow.

Strong point. As long as it does not cause too many scripts to fail or pick another app than before.

DnlLrssn:
> 
> 
> > I think it should navigate by default so that default matches the normal user flow.
> 
> Strong point. As long as it does not cause too many scripts to fail or pick another app than before.

If we add the apps from the space we move to (which we should) it'd cause scripts randomizing a selection from apps to also pick from these in actions subsequent to this one. So if navigating is default there need to be clear communication about this change (and bump minor on release of it).

SeBBBe:
Or maybe even go 1.0.0 😲

SeBBBe:
Yeah, it's a semi-breaking change. Not functionally, but the scenarios may become different. But we should probably do it anyway if it's a new default user flow in order to get away from legacy behavior as much as possible. That's my 2 cents at least.

atluq:
Artifact map logic is implemented, so this is ready for a second round of review

DnlLrssn (approved):

SeBBBe (approved):

PR #189, opened 2021-02-18
Add support for grid charts

DnlLrssn:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Add support for getting data for `sn-grid-chart`.

**Info**

Saving tree data pages has been added in this PR, however two steps still left to be able to support `select` in grid chart

* Add support for hypercube tree mode in `getpossible` part of select function.
* Add support for sending `SelectPivotCells` select request using the saved tree data pages.

The `allvalues` part of getting data might need re-evaluation in case we add more objects based on tree data in the future. Current will send correct requests for grid charts.

### DIFF
--- a/session/objecthandling.go
+++ b/session/objecthandling.go
@@ -608,9 +608,13 @@ func UpdateObjectHyperCubeTreeDataAsync(sessionState *State, actionState *action
 			nodes = append(nodes, node)
 		}
 
-		gob.GetHyperCubeTreeData(ctx, requestDef.Path, &enigma.NxTreeDataOption{
+		// TODO save tree to object hypercube?
+		_, err := gob.GetHyperCubeTreeData(ctx, requestDef.Path, &enigma.NxTreeDataOption{
 			TreeNodes: nodes,
 		})
+		if err != nil {
+			return errors.WithStack(err)
+		}
 
 		return nil
 	}, actionState, true, fmt.Sprintf("failed to get tree data for object<%s>", gob.GenericId))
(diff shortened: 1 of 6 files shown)


session/objecthandling.go
```diff
@@ -554,10 +560,66 @@ func UpdateObjectHyperCubeContinuousDataAsync(sessionState *State, actionState *
...
+
+		for i, dim := range dimInfo {
+			height := dim.Cardinal
+			if height > requestDef.Height {
+				height = requestDef.Height
+			}
+
+			allValues := true
+			if i == 0 {
+				allValues = false
```
SeBBBe, line 595: allValues := i != 0

SeBBBe (approved):
LGTM!

Is the lack of selection support the reason you were trying to take a look at the grid chart source code? If it's something that seems feasible to do we should open an issue on it for the future.

session/objecthandling.go
DnlLrssn: ah true, a bit cleaner

DnlLrssn: No wanted the code for more generic support for treedata and specifically `allValues` field. This is correct for grid chart which is the only such object right now.

For the select I know what to do now, it's just too much to change right now and want to prioritize other things.

SeBBBe: Perfect! That's fine; getting this level of support in the gopherciser is very useful not least for the app-evaluator. I agree that the select part is less important.

session/objecthandling.go
```diff
@@ -387,6 +389,7 @@ func UpdateObjectHyperCubeReducedDataAsync(sessionState *State, actionState *act
 	}, actionState, true, fmt.Sprintf("Failed to update object hypercube reduced data for object<%s>", gob.GenericId))
 }
 
+// UpdateObjectHyperCubeBinnedDataAsync sed get hypercube binned data request and update saved hypercube
```
atluq, line 392: sed --> send

atluq (approved):
LGTM!

session/objecthandling.go
DnlLrssn: `sed -i 's/sed/send/g' objecthandling.go`

PR #190, opened 2021-02-18
Add command line completion

atluq:
**Checklist**

- [] tests added
- [x] documentation updated

**Squash commit message**

Add command line completion

Add a cobra subcommand to generate command line completion scripts for zsh, bash, powershell and fish

**Info**

I have verified it works for zsh, bash and powershell. Any of you guys using fish?

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,0 +1,76 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	RootCmd.AddCommand(completionCmd)
+}
+
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate command line completion script",
+	Long: fmt.Sprintf(`To load completions:
+
+Bash:
+
+$ source <(%[1]s completion bash)
+
+# To load completions for each session, execute once:
+Linux:
+  $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
+MacOS:
+  $ %[1]s_osx completion bash > /usr/local/etc/bash_completion.d/%[1]s_osx
+
+Zsh:
+
+# If shell completion is not already enabled in your environment you will need
+# to enable it.  You can execute the following once:
+
+$ echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+# To load completions for each session, execute once:
+$ %[1]s completion zsh > "${fpath[1]}/_%[1]s"
+
+# You will need to start a new shell for this setup to take effect.
+
+Fish:
+
+$ %[1]s completion fish | source
+
+# To load completions for each session, execute once:
+$ %[1]s completion fish > ~/.config/fish/completions/%[1]s.fish
+
+Powershell:
+
+PS> %[1]s.exe completion powershell | Out-String | Invoke-Expression
+
+# To load completions for every new session, run:
+PS> %[1]s completion powershell > %[1]s.ps1
+# and source this file from your powershell profile.
+`, RootCmd.Name()),
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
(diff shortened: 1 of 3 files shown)


DnlLrssn:
> Any of you guys using fish?

not anymore, too many 3rd party scripts was written in way to not be compatible. Nice shell though.

atluq:
> > Any of you guys using fish?
> 
> not anymore, too many 3rd party scripts was written in way to not be compatible. Nice shell though.

Verified for fish too. It had some fancy defaults.

cmd/completion.go
```diff
@@ -0,0 +1,76 @@
...
+
+Zsh:
+
+# If shell completion is not already enabled in your environment you will need
+# to enable it.  You can execute the following once:
+
+$ echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+# To load completions for each session, execute once:
+$ %[1]s completion zsh > "${fpath[1]}/_%[1]s"
```
markuswall, line 37: $fpath[1] seems to point to `.oh-my-zsh/plugins/colored-man-pages` for me, I'm assuming that's not where the completions should end up.

cmd/execute.go
```diff
@@ -96,8 +96,8 @@ func (err SummaryTypeError) Error() string {
 var executeCmd = &cobra.Command{
 	Use:     "execute",
 	Aliases: []string{"x"},
-	Short:   "execute exerciser scenario towards a sense installation",
-	Long:    `execute exerciser scenario towards a sense installation`,
+	Short:   "Execute exerciser scenario towards a sense installation",
```
markuswall, line 99: Shouldn't this be _gopherciser_ scenario?

DnlLrssn: lol, yea probably, I think you found one of the oldest lines of code in the entire project, probably predates the naming of the repo... ;)

DnlLrssn: should probably also be changed to `a Sense environment` since there's so many different ones now, not only a windows installation.

markuswall:
Ok, I've tested quickly on WSL/zsh and cygwin/bash.
Seems to work well, mostly. Of some reason `--help` doesn't complete properly everywhere (e.g. `gopherciser execute --h<tab>` doesn't complete, but `gopherciser completion --h<tab>` does)

cmd/completion.go
```diff
@@ -0,0 +1,76 @@
...
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate command line completion script",
+	Long: fmt.Sprintf(`To load completions:
+
+Bash:
+
+$ source <(%[1]s completion bash)
+
+# To load completions for each session, execute once:
```
markuswall, line 23: These are not comments, thus no need for #'s.

atluq: Fixed

cmd/execute.go
atluq: Done

cmd/execute.go
```diff
@@ -96,8 +96,8 @@ func (err SummaryTypeError) Error() string {
 var executeCmd = &cobra.Command{
 	Use:     "execute",
 	Aliases: []string{"x"},
-	Short:   "Execute exerciser scenario towards a sense installation",
```
SeBBBe: 😆

cmd/completion.go
atluq: Added an extra comment in the help text on this. Cant come up with a simple standardized way here.

atluq:
> Ok, I've tested quickly on WSL/zsh and cygwin/bash.
> Seems to work well, mostly. Of some reason `--help` doesn't complete properly everywhere (e.g. `gopherciser execute --h<tab>` doesn't complete, but `gopherciser completion --h<tab>` does _in bash_)

The problem here is that the --help flag is created on first command execution. We could initialize all help flags at every execution, but dont know if it is worth it. Also tried to initialize all help flags when we run the completion command, but this seemed to be too late in the cobra setup flow.

markuswall (approved):
LGTM

PR #192, opened 2021-02-19
Apicompliance uploaddata

atluq:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Update uploaddata-action to follow the new request flow

**Info**

Where do we place the temp-file tus client?


scenario/uploaddata.go
```diff
@@ -70,124 +75,103 @@ func (settings UploadDataSettings) Execute(
 		return
 	}
 
-	restHandler := sessionState.Rest
+	sessionState.Rest.GetAsync(
+		fmt.Sprintf("%s/%s/quota", host, datafileEndpoint), actionState, sessionState.LogEntry, nil,
+	)
 
 	dataConnectionID, err := sessionState.FetchDataConnectionID(actionState, host, settings.SpaceID)
```
alfinhaji, line 82: `FetchDataConnectionID `needs to be updated to /api/v1/dc-dataconnections/DataFiles?type=connectionname

scenario/uploaddata.go
```diff
@@ -70,124 +75,103 @@ func (settings UploadDataSettings) Execute(
...
-	defer helpers.GlobalBufferPool.Put(body)
-	writer := multipart.NewWriter(body)
+	fileSize := len(fileContent)
+
+	tempContentPostRequest := sessionState.Rest.PostWithHeadersAsync(
+		fmt.Sprintf("%s/%s", host, tempContentFilesEndpoint),
+		actionState, sessionState.LogEntry, nil, map[string]string{
+			"upload-length": strconv.Itoa(fileSize),
+			"tus-resumable": "1.0.0",
+		}, &session.ReqOptions{
```
DnlLrssn, line 116: this looks like it should be uploaded using TUS? I.e. being able to automatically resume the upload on errors midstream etc. Have  a look at upload app where we use a library for TUS uploads.

atluq: Most definitely. Checking out upload app 👍

SeBBBe (approved):

scenario/uploaddata.go
```diff
@@ -31,8 +35,98 @@ type (
...
+		URL string
+	}
+
+	TempFileClient struct {
+		tusClient  *tus.Client
+		maxRetries int
+	}
+)
+
+func NewTempFileTUSClient(sessionState *session.State, connection *connection.ConnectionSettings, chunkSize int64, maxRetries int) (*TempFileClient, error) {
```
DnlLrssn, line 54: Maybe better to put in it's own files, in a helpers package or own package. Especially since it's used by more than one action.

DnlLrssn: otherwise lgtm, have not verified traffic though

atluq: Yes, I'll put it in a new package.

scenario/uploaddata.go
```diff
@@ -108,86 +102,47 @@ func (settings UploadDataSettings) Execute(
...
-	// Then create the binary multipart field
-	part, err := writer.CreateFormFile("data", filepath.Base(settings.Filename))
+	tempFileClient, err := tempcontent.NewTUSClient(sessionState, connection, defaultChunkSize, 5)
 	if err != nil {
-		actionState.AddErrors(errors.Wrapf(err, "failed to create multipart form"))
+		actionState.AddErrors(errors.WithStack(err))
 		return
 	}
-	_, err = io.Copy(part, file)
+	tempFile, err := tempFileClient.UploadFromFile(sessionState.BaseContext(), file)
```
DnlLrssn, line 111: Since I guess total upload time could take some time I understand why we don't use the new context function which would have the timeout set by `timeout` script parameter, but also not sure we should try uploading until failed or entire test times out. Maybe add an optional upload timeout flag (default set high like 1h our so), and create context  with this? This would be different from the retrybackoff.

DnlLrssn: not sure, opinions?

alfinhaji (approved):

docs/settingup.md
```diff
@@ -219,6 +219,7 @@ Apply a bookmark in the current app.
 
 * `title`: Name of the bookmark (supports the use of [variables](#session_variables)).
 * `id`: ID of the bookmark.
+* `selectionsonly`: *Missing documentation*
```
SeBBBe, line 222: Missing documentation

atluq: Old stuff, but I'm adding that

scenario/uploaddata.go
atluq: Added an opt-in timeout to uploaddata and uploadapp

PR #199, opened 2021-03-02
Apicompliance: elasticcreateapp

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Make elasticcreateapp action api compliant

**Info**

### DIFF
--- a/eventws/eventhandler.go
+++ b/eventws/eventhandler.go
@@ -91,7 +91,7 @@ func (handler *EventHandler) RegisterFunc(operation string, f func(event Event),
 }
 
 // RegisterFuncUntilCtxDone to be executed on event triggering. The function is
-// deregestered when context is done.
+// deregistered when context is done.
 // replay: triggers the latest events from buffer upon creation
 func (handler *EventHandler) RegisterFuncUntilCtxDone(ctx context.Context, operations []string, replay bool, f func(event Event)) {
 	for _, op := range operations {
--- a/eventws/eventhandler.go
+++ b/eventws/eventhandler.go
@@ -1,6 +1,7 @@
 package eventws
 
 import (
+	"context"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


eventws/eventhandler.go
```diff
@@ -89,6 +90,19 @@ func (handler *EventHandler) RegisterFunc(operation string, f func(event Event),
 	return eventFunc
 }
 
+// RegisterFuncUntilCtxDone to be executed on event triggering. The function is
+// deregestered when context is done.
+// replay: triggers the latest events from buffer upon creation
+func (handler *EventHandler) RegisterFuncUntilCtxDone(ctx context.Context, operations []string, replay bool, f func(event Event)) {
```
SeBBBe, line 96: Seems useful! 👍

SeBBBe (approved):
These kinds of PRs would be easier to review if there was a link to an issue or conversation about the required client changes. This one is straight-forward but it might not always be clear.

scenario/elasticcreateapp.go
```diff
@@ -34,6 +37,24 @@ func (settings ElasticCreateAppSettings) Execute(sessionState *session.State, ac
...
 	}
 
+	if !settings.IgnoreEvents {
+		events := sessionState.EventWebsocket()
+		if events == nil {
+			actionState.AddErrors(errors.New("Could not get events websocket"))
+			return
+		}
+
+		ctx, cancel := context.WithCancel(sessionState.BaseContext())
```
DnlLrssn, line 47: Should probably add a large default (and definable) timeout so we don't wait in this action forever if event never arrives. (same as should be added to other actions waiting for events). In this case maybe even just using the `timeout` setting from settings (i.e. the other context function from sessionstate) since this event should always be rather quick (other event waiters should use their own settings i.e. reload could take very long). If not received event before timeout -> report error

DnlLrssn (approved):

PR #200, opened 2021-03-04
log feature flag info row

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add info row log with current feature flags.

**Info**

### DIFF
--- a/session/features.go
+++ b/session/features.go
@@ -24,8 +24,8 @@ type (
 )
 
 var (
-	// FeatureFlagsLogged Global setting if feature flags has been logged
-	FeatureFlagsLogged bool
+	// LogFeatureFlags logs feature flags unless previously logged
+	LogFeatureFlags = &sync.Once{}
 )
 
 func (err FeatureAllocationError) Error() string {
@@ -55,10 +55,9 @@ func (features *Features) UpdateFeatureMap(rest *RestHandler, host string, actio
 		return
 	}
 
-	if !FeatureFlagsLogged {
-		FeatureFlagsLogged = true
+	LogFeatureFlags.Do(func() {
 		logEntry.LogInfo("FeatureFlags", fmt.Sprintf("%v", features.m))
-	}
+	})
 }
 
 // IsFeatureEnabled check if feature flag is enabled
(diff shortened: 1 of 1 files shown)


SeBBBe (approved):
How often will it run?

DnlLrssn:
> 
> 
> How often will it run?

once per openhub action. I.e. most likely once per user simulated for most scripts (unless multiple openhub actions in script)

DnlLrssn:
> 
> 
> How often will it run?

and now, only once

atluq (approved):
Looks good 👍

Could be an idea to use sync.Once, but the FeatureFlagsLogged bool is already synchronized by the mutex in Features.

DnlLrssn:
> 
> 
> Looks good 👍
> 
> Could be an idea to use sync.Once, but the FeatureFlagsLogged bool is already synchronized by the mutex in Features.

sure, not super important if it's truly once though, just that we don't spam it when haven 10's of k's of users.

SeBBBe:
I think any performance impact of sync.Once is negligible once it's been run once, as it's basically just checking an atomic boolean. Actually like the idea suggested by Anton to use it in case we have many users with each opening the hub.

> sync.Once() has an "optimized", short path that uses only an atomic load to check if the task has been called before, and if so, no mutex is used.

https://stackoverflow.com/questions/53376237/efficiency-measurments-of-gos-once-type

DnlLrssn:
> 
> 
> I think any performance impact of sync.Once is negligible once it's been run once, as it's basically just checking an atomic boolean. Actually like the idea suggested by Anton to use it in case we have many users with each opening the hub.
> 
> > sync.Once() has an "optimized", short path that uses only an atomic load to check if the task has been called before, and if so, no mutex is used.
> 
> https://stackoverflow.com/questions/53376237/efficiency-measurments-of-gos-once-type

Don't think we can expect that many users to hit this part of the code all at once, and if a few do, I see logging a few rows with same info as not an issue. I can add it, just saw it as a bad practice in principle to add unnecessary sync between separate users.

PR #202, opened 2021-03-04
Enhancement/755 event wait timeouts

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add possible timeout option to elasticreload action
* Don't log errors in elasticcreate app if base context is triggered

**Info**

Should we default to a large timeout instead of no timeout? E.g. maybe 1 hour?

### DIFF
--- a/scenario/elasticcreateapp.go
+++ b/scenario/elasticcreateapp.go
@@ -10,6 +10,7 @@ import (
 	"github.com/qlik-oss/gopherciser/connection"
 	"github.com/qlik-oss/gopherciser/elasticstructs"
 	"github.com/qlik-oss/gopherciser/eventws"
+	"github.com/qlik-oss/gopherciser/helpers"
 	"github.com/qlik-oss/gopherciser/session"
 )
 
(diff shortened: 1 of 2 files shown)


scenario/elasticreload.go
```diff
@@ -158,10 +164,18 @@ func (settings ElasticReloadSettings) execute(sessionState *session.State, actio
...
+	if settings.Timeout > 0 {
+		timeoutChan = time.After(time.Duration(settings.Timeout))
+	}
+
 	reloadID := postReloadResponse.ID
 forLoop:
 	for {
 		select {
+		case <-timeoutChan:
+			actionState.AddErrors(errors.New("timeout waiting on reload result event"))
```
SeBBBe, line 177: Thinking maybe we should call `checkStatusOngoing` or otherwise print the last known state when this timeout is reached? That way it's easier to see if something is messed up with the events or if it's just taking a really long time.

DnlLrssn: hmm good point.

scenario/elasticcreateapp.go
```diff
@@ -159,7 +160,7 @@ func (settings ElasticCreateAppSettings) Execute(sessionState *session.State, ac
 				return e.Operation == eventws.OperationCreated && e.ResourceID == appItemID
 			},
 		)
-		if err != nil {
+		if err != nil && !helpers.IsContextTriggered(sessionState.BaseContext()) {
```
atluq, line 163: This error will always be context triggered 🤔

SeBBBe (approved):

DnlLrssn:
Any opinions on optional (current implementation) or enforced timeout?

SeBBBe:
I would put a default timeout if none specified. Maybe an hour or so.

atluq (approved):
Looks good!

Default timeout or not does not really matter to me.

PR #206, opened 2021-03-10
Add select option for static selection to select action.

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

* Add `values` option to `select` action for statically selecting element values
* Remove action `staticselect`.

**Info**


SeBBBe (commented):
Is it really a good name to call it values? It's the element indices, right?

DnlLrssn:
same option functionality in QSST is also called `values` btw.

SeBBBe (approved):

atluq (approved):
LGTM! Makes sense to have static selection as a mode in select.

PR #217, opened 2021-03-26
update action interface to be able to return warnings on Validate function

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Changed `ActionSettings` interface to allow for script validation warnings being printed on execution
from:
```golang
ActionSettings interface {
		// Execute action
		Execute(sessionState *session.State, actionState *action.State, connectionSettings *connection.ConnectionSettings, label string /* Label */, reset func() /* reset action start */)
		// Validate action
		Validate() error
	}
```
to:
```golang
ActionSettings interface {
		// Execute action
		Execute(sessionState *session.State, actionState *action.State, connectionSettings *connection.ConnectionSettings, label string /* Label */, reset func() /* reset action start */)
		// Validate action []string are validation warnings to be reported to user
		Validate() ([]string, error)
	}
```

*  Changed `IScheduler` interface to allow for script validation warnings.
from:

```golang
	IScheduler interface {
		Validate() error
		...
```

to:

```golang
	IScheduler interface {
		Validate() ([]string, error)
		...
```

```bash
╭─dlo at SELUN-DLO1 in ~/dev/qlik-trial/gopherciser on enhancement/466-ValidateWarnings✘✘✘
╰─± ./build/gopherciser x -c ./localtests/perf4-videolink.json
2 script validation warnings:
1. warning 1
2. warning2
2021-03-26T08:41:44+01:00 Err<0> Warn<0> ActvSess<0> TotSess<0> Actns<0> Reqs<0>
```

**Info**

Currently change of interface only, no actual warnings added in this PR.

### DIFF
--- a/cmd/execute.go
+++ b/cmd/execute.go
@@ -187,7 +187,7 @@ func execute() error {
 		return JSONParseError(errUnmarshal.Error())
 	}
 
-	if err := cfg.Validate(); err != nil {
+	if err := validateConfigAndPrintWarnings(cfg); err != nil {
 		return JSONValidateError(err.Error())
 	}
 
--- a/cmd/script.go
+++ b/cmd/script.go
@@ -183,18 +183,18 @@ Will save one .structure file per app in script in the folder defined by output 
 }
 
 func validateConfigAndPrintWarnings(cfg *config.Config) error {
-	warnings, err := cfg.Validate()
+	err := cfg.Validate()
 	if err != nil {
 		return err
 	}
 
-	warningsCount := len(warnings)
+	warningsCount := len(cfg.ValidationWarnings)
 	if warningsCount < 1 {
 		return nil
 	}
 
 	_, _ = fmt.Fprintf(os.Stderr, "%d script validation warnings:\n", warningsCount)
-	for i, warning := range warnings {
+	for i, warning := range cfg.ValidationWarnings {
 		_, _ = fmt.Fprintf(os.Stderr, "%d. %s\n", i+1, warning)
 		if i == 9 {
 			_, _ = fmt.Fprintf(os.Stderr, "...(%d) additional warnings\n", warningsCount-i+1)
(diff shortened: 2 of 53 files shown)


SeBBBe (commented):
Looks good!

One idea though - perhaps we should also log the warnings to the log. I can foresee a scenario where someone ignores warnings and something weird happens, and there may or may not arise a situation of incessant asking about script warnings :) So if it's in the log it'll be easier for everyone.

DnlLrssn:
> 
> 
> Looks good!
> 
> One idea though - perhaps we should also log the warnings to the log. I can foresee a scenario where someone ignores warnings and something weird happens, and there may or may not arise a situation of incessant asking about script warnings :) So if it's in the log it'll be easier for everyone.

Except the log has not been created at that point I think?

SeBBBe:
> > Looks good!
> > One idea though - perhaps we should also log the warnings to the log. I can foresee a scenario where someone ignores warnings and something weird happens, and there may or may not arise a situation of incessant asking about script warnings :) So if it's in the log it'll be easier for everyone.
> 
> Except the log has not been created at that point I think?

Maybe the warnings array could be saved somewhere and then logged once the log is initialized?

DnlLrssn:
> 
> 
> > > Looks good!
> > > One idea though - perhaps we should also log the warnings to the log. I can foresee a scenario where someone ignores warnings and something weird happens, and there may or may not arise a situation of incessant asking about script warnings :) So if it's in the log it'll be easier for everyone.
> > 
> > 
> > Except the log has not been created at that point I think?
> 
> Maybe the warnings array could be saved somewhere and then logged once the log is initialized?

Yea I think something like that would be only option, just verified, logging is not setup until `Execute`.

Realized we probably should change scheduler interface so added that too.

DnlLrssn:
Added to log on `info` level with info type `ScriptValidationWarning`. could add on warning level, but felt better to keep warnings to warnings during test, right?

SeBBBe (approved):
Perfect!

atluq (approved):

PR #218, opened 2021-04-01
Fix unmarshaling bug in QSEST

SeBBBe:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

**Info**

Quite crude fix, but it's in line with how we handle it in other actions. Might pose a problem if someone actually want to execute load tests under a custom compile under js but unlikely scenario.. And that wouldn't work anyway.

Also removed some old documentation that's since been moved out of this repo.


DnlLrssn (changes requested):
Why have the data files for the doc compile test case been deleted?

SeBBBe:
Right, it's for the test case.. It's confusing. Should we really keep references to internal actions even if it's a test case though? I either fix up the test cases or revert the removal of the data.

scenario/askhubadvisor.go
```diff
@@ -164,12 +164,11 @@ func (settings *AskHubAdvisorSettings) UnmarshalJSON(bytes []byte) error {
...
 	case QueryFromFile:
-
 		if settings.FileName == "" {
 			return errors.New("no file name")
 		}
 		if runtime.GOOS == "js" {
-			return errors.Errorf("can not read file with GOOS=%s", runtime.GOOS)
+			return nil
 		}
 		file, err := os.Open(settings.FileName)
```
atluq, line 173: Shouldn't this be moved to execute?

atluq: `os.Open` I mean

atluq:
> 
> 
> Right, it's for the test case.. It's confusing. Should we really keep references to internal actions even if it's a test case though? I either fix up the test cases or revert the removal of the data.

Right, since it is a test case it is ok either way. I mean they are still in the git history.

scenario/askhubadvisor.go
SeBBBe: If it's in Execute it will be executed for each user though? I guess there may be some way around that but it doesn't really make a difference. Having it in Umarshal is fine as long as it doesn't run in js.

atluq (approved):

DnlLrssn (approved):

PR #224, opened 2021-04-15
Add configurable askhubadvisor followup type

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Add configurable askhubadvisor followup type

The hubadvisor responses contain info on different types of followups. This change adds the possiblity
to filter the followups which are actually performed in the script,  using this info. The primary use case is 
to limit the action to select an app where followup is asked.


scenario/askhubadvisor.go
```diff
@@ -71,6 +71,7 @@ type (
 		SaveImages        bool                   `json:"saveimages" displayname:"Save images" doc-key:"askhubadvisor.saveimages"`
 		SaveImageFile     session.SyncedTemplate `json:"saveimagefile" displayname:"File name (without suffix)" doc-key:"askhubadvisor.saveimagefile" displayelement:"savefile"`
 		ThinkTimeSettings *ThinkTimeSettings     `json:"thinktime,omitempty" displayname:"Think time settings" doc-key:"askhubadvisor.thinktime"`
+		FollowupTypes     []followupType         `json:"followuptypes,omitempty" displayname:"Followup query type priority" doc-key:"askhubadvisor.followuptypes"`
```
SeBBBe, line 74: Not sure how this will play with QSEST? An array of a custom type

scenario/askhubadvisor.go
```diff
@@ -157,6 +164,40 @@ type (
...
+
+	followupRecommendation followupType = "recommendation"
+	followupMeasure        followupType = "measure"
+	followupApp            followupType = "app"
+	followupDimension      followupType = "dimension"
+	followupSentence       followupType = "sentence"
+)
+
+var (
+	validFollowupTypes = map[followupType]struct{}{
```
SeBBBe, line 181: Why not use an enummap?

DnlLrssn (commented):
Other than possible issues with GUI as @SeBBBe mentioned, looks good to me

scenario/askhubadvisor.go
atluq: Is an array of enummaps ok with the GUI?

SeBBBe: No, it will require some kind of new form.. I guess equivalent to current logic for strings or numbers, but with dropdowns for selecting the value.

SeBBBe (approved):
Amazing!

PR #225, opened 2021-04-16
Adds support for using custom session variables in scenario scripts.

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

* Adds support for using custom session variables in scenario scripts.
* Change default Go version to 1.16 and minimal version to 1.14

**Info**


docs/settingup.md
```diff
@@ -1720,6 +1722,25 @@ The following functions are supported:
...
+        "name": "BookmarkCounter",
+        "type": "int",
+        "value": "1"
+    }
+},
+{
+  "action": "createbookmark",
+  "settings": {
+    "title": "Bookmark no {{ add .ScriptVars.BookmarkCounter 1 }}",
+    "description": "This bookmark wil have the title Bookmark no 2"
```
SeBBBe, line 1739: will*

generatedocs/data/actions/setscriptvar/examples.md
```diff
@@ -0,0 +1,136 @@
...
+{
+    "action": "iterated",
+    "disabled": false,
+    "settings": {
+        "iterations": 3,
+        "actions": [
+            {
+                "action": "createbookmark",
+                "settings": {
+                    "title": "Bookmark {{ index .ScriptVars.MyArrayVar .ScriptVars.BookmarkCounter }}",
```
SeBBBe, line 121: Pretty cool stuff!

generatedocs/data/params.json
```diff
@@ -447,6 +447,18 @@
     "setscript.script": [
         "Load script for the app (written as a string)."
     ],
+    "setscriptvar.name" : [
+        "Name of variable to set. Will overwrite any existing variable with same name."
+    ],
+    "setscriptvar.type" : [
+        "Type of the variable.",
+        "`string`: Variable of type string e.g. `my var value`.",
+        "`int`: Variable of type integer e.g. `6`.",
```
SeBBBe, line 456: No need for floats?

session/syncedTemplate.go
```diff
@@ -125,3 +127,39 @@ func (input *SyncedTemplate) ReplaceWithoutSessionVariables(data interface{}) (s
...
+
+func parseToInt64(val interface{}) (int64, error) {
+	switch val := val.(type) {
+	case string:
+		val2, err := strconv.ParseInt(val, 10, 64)
+		if err != nil {
+			return 0, errors.Wrapf(err, "string<%s> not parseable to integer", val)
+		}
+		return val2, nil
+	case int:
```
SeBBBe, line 152: Use a case list to avoid repeating the same line

SeBBBe (approved):
Nice! Simple implementation and powerful. Some minor things to fix.

generatedocs/data/params.json
DnlLrssn: was thinking of it, do we? as a specific type that is.

session/syncedTemplate.go
DnlLrssn: not possible due to the case to int64

DnlLrssn: e.g. this is *not*  valid syntax due to the cast.
```golang
	case int, int8:
		return int64(val), nil

```

docs/settingup.md
DnlLrssn: fixed

session/syncedTemplate.go
DnlLrssn: This is due to although they look the same to us, they don't look the same to the compiler, it'll see it as
```golang
case int:
   return int64(int)
 case int8, int16:
   return int64(interface{})
```

SeBBBe: 🤯

generatedocs/data/params.json
SeBBBe: Don't know tbh. Just seems like it's missing. Probably someone will request it at some point, along with a multiplication/division operator..

DnlLrssn: mm and will get messier with those kind of functions, so thought I'd keep it simple for now until someone request it with an actual case. You think we should add it already now?

SeBBBe: Meh let's keep it this way for now then

DnlLrssn: also first felt it was missing, but change my mind to start simpler. Have no problem of adding it later if there's a request for it.

generatedocs/data/actions/setscriptvar/examples.md
atluq: Indeed, good example!

Is it possible to do push and pop operations on array using setscriptvar and template stuff?

atluq (approved):
Cool stuff!

generatedocs/data/actions/setscriptvar/examples.md
DnlLrssn: Hmm no, since sescriptvar right recreates the variable, will think a bit about this

DnlLrssn: you can append 

```json
{
    "action": "setscriptvar",
    "settings": {
        "name": "MyVarArray",
        "type": "array",
        "value": "val1,val2,val3",
        "sep": ","
    }
},
{
    "action": "setscriptvar",
    "settings": {
        "name": "MyVarArray",
        "type": "array",
        "value": "{{ range $i,$element:=(split .MyArray ",") }}{{if ne 0 $i}},{{ end }}{{$element}}{{ end }},val4",
        "sep": ","
    }
}
```
should produce an array with val1,val2,val3,val4

so possible but currently not nice
 {{ range $i,$element:=(split .MyArray ",") }}{{if ne 0 $i}},{{ end }}{{$element}}{{ end }},val4

DnlLrssn: could make it nicer by adding a join function

DnlLrssn: although pop can technically be done with `slice {{ slice .ScriptVars.MyArray 0 (add (len .ScriptVars.MyArray) -1) }}` we have two problems here, 
1. If the result is "", we will fail on the creation of the variable since we test towards ""
2. len .ScriptVars.MyArray will not evaluate to 0 on nil (as when done in code) and instead through an error

So will add the `join` function and example how to create an array adding to it when creating bookmarks, then iterate it to delete them

DnlLrssn: @atluq , ok added the `join` function and made the more advanced scenario using it. Not exactly push/pop, but don't want to add functions that actually manipulates the value of the variables in the templates.

atluq: Great, taking a look. Totally with you on no variable mutation in template functions.

atluq (approved):
Great examples and addition of join-function!

.circleci/config.yml
```diff
@@ -123,7 +123,7 @@ jobs:
   build-publish-docker:
     <<: *defaults
     docker:
-      - image: circleci/golang:1.15
+      - image: circleci/golang:1.16
```
SeBBBe, line 126: 🚀

docs/settingup.md
```diff
@@ -1331,6 +1331,242 @@ Set the load script for the current app. To load the data from the script, use t
...
+
+Sets a variable which can be used within the same session. Cannot be accessed across different simulated users.
+
+* `name`: Name of variable to set. Will overwrite any existing variable with same name.
+* `type`: Type of the variable.
+    * `string`: Variable of type string e.g. `my var value`.
+    * `int`: Variable of type integer e.g. `6`.
+    * `array`: Variable of type arrat e.g. `[1, 2, 3]`.
+* `value`: Value to set to variable (supports the use of [session variables](#session_variables)).
+* `sep`: *Missing documentation*
```
SeBBBe, line 1347: Add documentation

SeBBBe (approved):
Looks great! Good examples!

docs/settingup.md
DnlLrssn: oh, missed that one

DnlLrssn: fixed

SeBBBe (approved):

PR #232, opened 2021-05-06
Add configurable PRE and POST execution hooks.

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

* Moves synced templates to their own package
* Adds pre and post execution hooks which can be used to send requests with data before and after test is executed.
* Add floats NearlyEqual and NearlyEqualEpsilon helper functions.

**Info**


config/hooks.go
```diff
@@ -0,0 +1,445 @@
...
+	req, err := http.NewRequestWithContext(ctx, strings.ToUpper(httpMethodEnum.StringDefault(int(hook.Method), http.MethodPost)), hook.Url, buf)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	req.Header.Set("Content-Type", hook.ContentType)
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+
+	// TODO try avoid creating client here
```
SeBBBe, line 302: It will only be done once pre and once post test though, right? So should be a small overhead

SeBBBe (approved):
Will be some work if we want to implement it in the GUI but otherwise looks great! I verified that opening and saving a script without affecting the hook settings work though so we're all good from that point of view.

config/hooks.go
DnlLrssn: Correct. But if we need to change any of the settings at some point we might forget to change this one too.

DnlLrssn:
> Will be some work if we want to implement it in the GUI 

ok any part in particular that doesn't work out of the box?

SeBBBe:
> > Will be some work if we want to implement it in the GUI
> 
> ok any part in particular that doesn't work out of the box?

All the root level parts require some special handling to appear so can't say for sure. All the components for strings, arrays etc are automatically reused but the base level expander needs to be manually inserted in one of the panes.

The part I worry about the most would be the extractors slice but I can have a closer look and see what happens.

SeBBBe:
Screenshot 2021-05-06 at 13 21 01

SeBBBe:
It mostly works.
* Extractors can't be created from GUI
* There's no distinction between which is the pre and post hook:

Screenshot 2021-05-06 at 13 22 03

SeBBBe:
If I create a new script and try to save it immediately, I get this validation error:

`config.HookCore.Content: Method: unmarshalerDecoder: Failed to unmarshal httpMethodEnum: Unknown enum<>: Key<> not found, error found in #10 byte of ...|method":"","payload"|..., bigger context ...|{"contenttype":"","extractors":[],"method":"","payload":"","respcodes":[],"url":""}|...`

SeBBBe:
Seems like creating extractors fails because this:

sliceElemType := reflect.TypeOf(input).Elem()
sampleValue := reflect.New(sliceElemType) // Instantiate new member of slice element type
sample := sampleValue.Elem().Interface()
inputType := reflect.TypeOf(input)

Ends up with inputType == nil

helpers/floats.go
```diff
@@ -0,0 +1,32 @@
...
+
+// NearlyEqual float points covering some edge cases as per https://floating-point-gui.de/errors/comparison/#look-out-for-edge-cases
+func NearlyEqualEpsilon(a, b float64, epsilon float64) bool {
+	absA := math.Abs(a)
+	absB := math.Abs(b)
+	diff := math.Abs(a - b)
+
+	if a == b { // shortcut, handles infinities
+		return true
+	} else if a == 0 || b == 0 || diff < FloatMinNormal {
```
atluq, line 25: `diff` shall be `absA + absB`

atluq (commented):
Looks good!

Just a typo in the floating point comparison implementation.

helpers/floats.go
DnlLrssn: hmm interesting, actually converted the code the shows up in duckduckgo which is from stackoverflow (but wanted to link to original author), comparing what's on the site and what shows there is different, must have been updated
image

DnlLrssn: good find. will update.

SeBBBe (approved):

SeBBBe:
After some adjustments by Daniel and to QSEST by me, it's now fully working in the GUI albeit perhaps not the prettiest.

Screenshot 2021-05-07 at 13 10 32

SeBBBe:
Seems like the current way of creating a default is not acceptable:

```
"hooks": {
        "preexecute": {
            "url": "",
            "method": "get",
            "payload": "",
            "respcodes": [
                200,
                201
            ],
            "contenttype": "application/json",
            "extractors": [],
            "headers": null
        },
        "postexecute": {
            "url": "",
            "method": "get",
            "payload": "",
            "respcodes": [
                200,
                201
            ],
            "contenttype": "application/json",
            "extractors": [],
            "headers": null
        }
    },
```

DnlLrssn:
@SeBBBe hook headers should now show up in GUI better I hope?

SeBBBe:
> @SeBBBe hook headers should now show up in GUI better I hope?

It works perfectly in the GUI! Will open a PR with the changes I already made to sort them out as a separate expander.

PR #240, opened 2021-06-07
add lookup functions for item ID to artifactmap

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add artifactmap lookup functions also for item ID.

**Info**

### DIFF
--- a/session/artifactmap.go
+++ b/session/artifactmap.go
@@ -44,6 +44,8 @@ type (
 		Title string
 		Type  string
 	}
+
+	ArtifactEntryCompareType int
 )
 
 // Currently known resource types
@@ -52,6 +54,12 @@ const (
 	ResourceTypeGenericLink = "genericlink"
 	ResourceTypeDataset     = "dataset"
 	ResourceTypeDataAsset   = "dataasset"
+)
+
+const (
+	ArtifactEntryCompareTypeID ArtifactEntryCompareType = iota
+	ArtifactEntryCompareTypeItemID
+	ArtifactEntryCompareTypeName
 )
 
 func (d *ArtifactList) Len() int {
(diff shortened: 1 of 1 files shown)


session/artifactmap.go
```diff
@@ -135,24 +143,29 @@ func (am *ArtifactMap) FillArtifacts(item *ItemData) error {
 
 // DeleteApp deletes an app from the ArtifactMap
 func (am *ArtifactMap) DeleteApp(appGUID string) {
-	am.DeleteItem(ResourceTypeApp, appGUID, true)
+	am.DeleteItem(ResourceTypeApp, appGUID, ArtifactEntryCompareTypeID)
```
SeBBBe, line 146: Just incorrect variable naming? Appid != itemid

DnlLrssn: AppID is saved in entry.ID
ItemID is saved in entry.ItemID
so correct variable naming as far as I can see

SeBBBe: ArtifactEntryCompareTypeID implies item Id?

SeBBBe: Alright. Well, .ID is used in most other places to signify itemId so I guess that's why I thought it seemed off.

SeBBBe (approved):

session/artifactmap.go
DnlLrssn: well previous code was
```
Lookup(resourcetype, lookup string, id bool)
```
where `id` mean app GUID (in this case)

DnlLrssn: entry.ID as artifactmap entry is always resource ID not item ID

DnlLrssn:
```
	ArtifactEntry struct {
		Name         string      `json:"name"`
		ID           string      `json:"resourceId"`
		ItemID       string      `json:"ID,omitempty"`

```

SeBBBe: That snippet shows the confusion very well 😆

DnlLrssn: yea noticed the json say ID for some reason lol... well should only become json for logs so should be ok

SeBBBe: The way I see it, the json is the way most places refer to the itemid vs resourceid and the internal variable names are what is used in the gopherciser

DnlLrssn: It used to be the resourceid only, the item id was a later addition.

PR #242, opened 2021-06-09
Refactored push metrics URL so it supports path on URL as well

HvonP:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Refactored push metrics URL so it supports path on URL as well

**Info**

Refactored push metrics URL so it supports path on URL as well

### DIFF
--- a/metrics/transport.go
+++ b/metrics/transport.go
@@ -91,7 +91,7 @@ func PushMetrics(ctx context.Context, metricsPort int, metricsAddress, job strin
 
 	u, err := url.Parse(metricsAddress)
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("can't parse metricsAddress <%s>, metrics will not be pushed", metricsAddress)
 	}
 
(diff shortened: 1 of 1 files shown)


metrics/transport.go
```diff
@@ -88,7 +89,12 @@ func PushMetrics(ctx context.Context, metricsPort int, metricsAddress, job strin
 		return err
 	}
 
-	var addr = flag.String("push-address", fmt.Sprintf("%s:%d", metricsAddress, metricsPort), "The address to push prometheus metrics")
+	u, err := url.Parse(metricsAddress)
+	if err != nil {
+		panic(err)
```
DnlLrssn, line 94: Is this the behaviour we want? If we can't parse a push gateway (and thus send metrics) don't execute the test at all?

HvonP: Altered the code to print and error (stderr) and ignore trying to push metrics

DnlLrssn (approved):

ashishkumarqlik (approved):
LGTM

PR #250, opened 2021-08-06
swith from github.com/dgrijalva/jwt-go to github.com/golang-jwt/jwt/v4

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Switch to use use different jwt ´package due to CVE-2020-26160. Issue not really affecting gopherciser but still better to change. enigma-go still uses the vulnerable package so no fully closed event with this change.

**Info**

reviewers chosen "at random" since no one else from dev team here


DnlLrssn:
hmm seems go1.14 doesn't support this change, will decide what to do with this next week.

pawel-fobka (approved):

PR #261, opened 2021-09-22
Change how execution cancel on maxerrors works

DnlLrssn:
Change how execution cancel on maxerrors works, due to not getting triggered in sub actions. Breaking change! since flag is moved to config.Settings instead of scheduler

**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

Breaking change. config.scheduler.maxerrors moved to config.settings.maxerrors.

Sub actions, such as actions created by the `randomaction` did not cancel execution when max errors was reached. Re-work this functionality to cover also this case.

**Info**


SeBBBe (approved):

atluq (approved):

foh79 (approved):
Lgtm maxerrers now work outside and inside randomactions

PR #272, opened 2021-11-02
add setproperties to createsheet

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Set two properties on sheet after creating it

**Info**

### DIFF
--- a/scenario/createsheet.go
+++ b/scenario/createsheet.go
@@ -79,6 +79,8 @@ func (settings CreateSheetSettings) Execute(sessionState *session.State,
 			if err != nil {
 				return err
 			}
+		} else if err != nil {
+			return err
 		}
 
 		_, err = jsonparser.GetString(prop, "layoutOptions")
@@ -87,6 +89,8 @@ func (settings CreateSheetSettings) Execute(sessionState *session.State,
 			if err != nil {
 				return err
 			}
+		} else if err != nil {
+			return err
 		}
 
 		if err := genObj.SetPropertiesRaw(ctx, prop); err != nil {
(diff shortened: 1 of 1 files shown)


ashishkumarqlik (approved):
lgtm

scenario/createsheet.go
```diff
@@ -66,6 +67,32 @@ func (settings CreateSheetSettings) Execute(sessionState *session.State,
 			return sessionState.IDMap.Add(settings.ID, genObj.GenericId, sessionState.LogEntry)
 		}
 
+		prop, err := genObj.GetPropertiesRaw(ctx)
+		if err != nil {
+			return err
+		}
+
+		// client updates two parameters, do this safely jsonparser as using a struct might "lose" values
```
SeBBBe, line 75: Interesting 🤔  We might create a helper for this if it becomes more commonplace to do this kind of thing, for now it's great like it is!

SeBBBe (approved):

scenario/createsheet.go
```diff
@@ -66,6 +67,32 @@ func (settings CreateSheetSettings) Execute(sessionState *session.State,
...
+		_, err = jsonparser.GetString(prop, "gridResolution")
+		if err == jsonparser.KeyPathNotFoundError {
+			prop, err = jsonparser.Set(prop, []byte(`"small"`), "gridResolution")
+			if err != nil {
+				return err
+			}
+		}
```
atluq, line 82: Should we return errors that are not `jsonparser.KeyPathNotFoundError`?

DnlLrssn: hmm yes, good point

DnlLrssn: fixed

PR #279, opened 2021-11-22
Better support calculation conditions in some charts

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Better support calculation conditions for charts with separate data requests.

**Info**

```bash
λ ~/dev/qlik-oss/gopherciser/ bugfix/954-CalcCondition* make quickbuild &&  ./build/gopherciser_osx x -c ./localtests/rd-ajh-openreloadanalyzer.json
GO111MODULE=on go build -mod=readonly -o  ./build/gopherciser_osx
2021-11-22T16:33:53+01:00 Err<0> Warn<0> ActvSess<0> TotSess<0> Actns<0> Reqs<0>
TotErrors<0> TotWarnings<1> TotActions<4> TotRequests<103> Duration<2.373589499s> 
λ ~/dev/qlik-oss/gopherciser/ bugfix/954-CalcCondition* grep -E "\twarning\t" ./logs/$(ls -t ./logs | head -n 1)                                    
2021-11-22T16:33:55.742584+01:00        changesheet             4       warning         object<BSpwem>: object has unsatisfied calculation condition                    0       Reload Analyzer v1.2.1-3        d19b9b59-983d-4723-9422-fe329c0108d7    dlotesters_111               11              0       0               0       0       0       2021-11-22T15:33:55.742584Z
```

### DIFF
--- a/session/objecthandling.go
+++ b/session/objecthandling.go
@@ -303,10 +303,7 @@ func UpdateObjectHyperCubeDataAsync(sessionState *State, actionState *action.Sta
 			switch err.(type) {
 			case CalcEvalConditionFailedError:
 				sessionState.LogEntry.Logf(logger.WarningLevel, "object<%s>: %v", obj.ID, err)
-				if err = obj.SetHyperCubeDataPages(make([]*enigma.NxDataPage, 0), false); err != nil {
-					return errors.Wrap(err, "failed to set hypercube datapages")
-				}
-				return nil
+				return errors.Wrap(obj.SetHyperCubeDataPages(make([]*enigma.NxDataPage, 0), false), "failed to set hypercube datapages")
 			}
 			return errors.WithStack(err)
 		}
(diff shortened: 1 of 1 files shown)


session/objecthandling.go
```diff
@@ -292,14 +299,22 @@ func UpdateObjectHyperCubeDataAsync(sessionState *State, actionState *action.Sta
...
-		if hypercube.Size == nil {
-			return errors.Errorf("object<%s> has no hypercube size", gob.GenericId)
-		}
-
 		if err := checkHyperCubeErr(gob.GenericId, hypercube.Error); err != nil {
+			switch err.(type) {
+			case CalcEvalConditionFailedError:
+				sessionState.LogEntry.Logf(logger.WarningLevel, "object<%s>: %v", obj.ID, err)
+				if err = obj.SetHyperCubeDataPages(make([]*enigma.NxDataPage, 0), false); err != nil {
+					return errors.Wrap(err, "failed to set hypercube datapages")
```
SeBBBe, line 307: You could just
> return errors.Wrap(err, "failed to set hypercube datapages")

unconditionally because the line after returns a nil. In this case err is nil so the wrap would return nil as well.

SeBBBe (approved):

session/objecthandling.go
DnlLrssn: yes true, just copy pasted the lines from the different set data parts, but makes sense

atluq (approved):

PR #280, opened 2021-12-03
Add qInExtRow as insignificant value for regression analysis

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add `qInExtRow` as insignificant value for regression analysis

**Info**

### DIFF
--- a/logger/regression.go
+++ b/logger/regression.go
@@ -41,6 +41,7 @@ var filters = marshalFilters(
 	"+qPivotDataPages",
 	"+qStackedDataPages",
 	"-qNum",
+	"-qInExtRow",
 )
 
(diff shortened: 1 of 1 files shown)


SeBBBe:
I tested it by adding the filter manually to the log Mounika produced.

Can confirm that the baseline log is the one where the filter is taken from.

SeBBBe:
Screenshot 2021-12-03 at 13 12 14

Screenshot 2021-12-03 at 13 12 07

SeBBBe:
Remaining are some changes in qText in a lot of objects

SeBBBe (approved):

PR #285, opened 2021-12-16
Performance improvements

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Performance improvements based on profiling results

results in 33% less CPU usage, minimal increase in RAM and almost 30% less system calls.

* Improvement: Change json encoder/decoder from json-iterator/go to github.com/goccy/go-json
* Improvement: Cache dns lookups done from REST handler
* Bugfix: REST handler incomming traffic logger allocating memory for Body when not needed.

**Info**


go.mod
```diff
@@ -8,16 +8,17 @@ require (
...
 	github.com/google/uuid v1.3.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/json-iterator/go v1.1.12
 	github.com/kr/text v0.2.0 // indirect
 	github.com/mitchellh/go-ps v1.0.0
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
-	github.com/qlik-oss/enigma-go/v3 v3.0.1
+	github.com/qlik-oss/enigma-go/v3 v3.0.2-0.20211216135847-b31ee494315e
```
SeBBBe, line 20: Integration version with new goccy/go-json?

SeBBBe (commented):
New error message format on script unmarshal errors?

go.mod
DnlLrssn: correct + Float64 unmarshal changes (and some updated dependencies)

DnlLrssn: yes new error message from the part that comes from the used library yes, and yes is less verbose

PR #292, opened 2022-01-17
Failsafe against hanging gopherciser after stop requested.

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add failsafe in case process is not killed within 5 minutes after cancelling main context. Create a file such as `gopherciser_12345_hang.stack` with all running go routine stacks before force quitting.
* Improve pending mark/unmark during eventing websocket reconnection.


atluq (approved):
Great approach to find out more about the hanging problem!

PR #295, opened 2022-01-26
Add SmartSearch action

atluq:
**Checklist**

- [x] tests added
- [x] documentation updated
- [x] Add selection from search results
- [x] Do not retry aborted ws requests

**Squash commit message**

Add SmartSearch action


scenario/smartsearch.go
```diff
@@ -0,0 +1,200 @@
...
+	doc := uplink.CurrentApp.Doc
+	searchTerms := settings.searchTerms
+	sessionState.LogEntry.LogDebugf(`search terms: "%s"`, strings.Join(searchTerms, `","`))
+
+	var searchResult *enigma.SearchResult
+	var searchSuggestionResult *enigma.SearchSuggestionResult
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
```
DnlLrssn, line 144: if there's a panic entire gopherciser (or other service using this as lib) would be taken down due to no panic recover. Try to use the async functions functions with callbacks instead followed by a sessionState.Wait (appliest to all the usuage of `go` directly in the file)

scenario/smartsearch.go
```diff
@@ -0,0 +1,200 @@
...
+		defer wg.Done()
+		err := sessionState.SendRequest(actionState, func(ctx context.Context) error {
+			var err error
+			searchResult, err = doc.SearchResults(
+				ctx,
+				&enigma.SearchCombinationOptions{
+					Context:      "CurrentSelections",
+					CharEncoding: "Utf16",
+				},
+				searchTerms,
```
SeBBBe, line 171: So this will always send all search times as one request?

docs/settingup.md
```diff
@@ -1821,6 +1821,76 @@ Create and execute a `changesheet` action for each sheet in an app. This can be
...
 </details>
 
+<details>
+<summary>smartsearch</summary>
+
+## SmartSearch action
+
+Perform a Smart Search in Sense app to find suggested selections.
+
+* `searchtext`: Text used for search. Search terms are separated by spaces. Use double quotes include spaces in search terms.
```
SeBBBe, line 1831: No biggie but would prefer having a comma as delim

SeBBBe (commented):
A couple of things prevent this from replacing existing tests.
* Ability to randomize search term - we need to be able to specify many search terms either in a file or inline and then have the action randomly choose one for each user
* No support for "slow typer". This is the case where a user doesn't past in a search query, but instead types in character by character with some delay in between. This will generate search and search suggestion calls for each character typed

DnlLrssn (changes requested):
This is equivalent of user pasting the the search query since only one search is sent. In i.e. same as the `fasttyper` : true setting in `sdkexerciser`. Normally behavior when a user writes the search query is that search would be sent multiple times.

Current implementation would not be very useful since all users would do the exact same search. We need a way to have lists with different searches from which a search can be randomized. Preferably also with a from file option.

atluq:
Alright, I'll fix random searchtext from file or similar. I did not experience multiple rpc calls when typing the search, but of course it can happen. How far shall we go with the typing simulation and mimicing of the typing debouncer in the browser client?

scenario/smartsearch.go
atluq: Yes, then you get a one list item each in the response for all combinations of terms.

DnlLrssn:
> Alright, I'll fix random searchtext from file or similar. I did not experience multiple rpc calls when typing the search, but of course it can happen. How far shall we go with the typing simulation and mimicing of the typing debouncer in the browser client?

ok, this might be an client change then, it use to send and update results while you where typing. Personally haven't used smart search for long time.

DnlLrssn:
ahh also missing an option to `select´ or not `select` the result (in sdkexerciser action `makeselection` setting)

atluq:
> > Alright, I'll fix random searchtext from file or similar. I did not experience multiple rpc calls when typing the search, but of course it can happen. How far shall we go with the typing simulation and mimicing of the typing debouncer in the browser client?
> 
> ok, this might be an client change then, it use to send and update results while you where typing. Personally haven't used smart search for long time.

tried it a little more and when I made errors and had to use backspace, some type of typing simulation would be necessary 🤔

DnlLrssn:
> > > Alright, I'll fix random searchtext from file or similar. I did not experience multiple rpc calls when typing the search, but of course it can happen. How far shall we go with the typing simulation and mimicing of the typing debouncer in the browser client?
> > 
> > 
> > ok, this might be an client change then, it use to send and update results while you where typing. Personally haven't used smart search for long time.
> 
> tried it a little more and when I made errors and had to use backspace, some type of typing simulation would be necessary 🤔

in sdkexerciser we used 240CPM in typer simulation. but seems client sent partial results more often back then.

DnlLrssn:
don't think we should simulate too much erasing/rewriting etc though.

scenario/smartsearch.go
atluq: Fixed

SeBBBe:
I vote for having a delay with some type of distribution or random pauses. For example do a 240 CPM with a 5% chance each character that there would be a couple seconds pause? I think something like that should be a reasonable realistic approximation of behavior.

scenario/smartsearch.go
```diff
@@ -125,6 +208,115 @@ func parseSearchTerms(s string) []string {
...
+	}
+
+	const typeOneCharDuration = 300 * time.Millisecond
+	const minPostTypingDuration = 700 * time.Millisecond
+	const maxPostTypingDuration = 1300 * time.Millisecond
+
+	chunks := searchTextChunks{}
+	currentStart := 0
+	for currentEnd := 1; currentEnd < len(searchText); currentEnd++ {
+		randInt, err := randomizer.RandWeightedInt([]int{1, 9})
```
SeBBBe, line 275: Would be better if this propbability was also a const

scenario/smartsearch.go
```diff
@@ -125,6 +208,115 @@ func parseSearchTerms(s string) []string {
...
+		return searchTextChunks{
+			searchTextChunk{
+				Text:                    searchText,
+				PostTypingThinkDuration: 0,
+				TypingDuration:          0,
+			},
+		}, nil
+	}
+
+	const typeOneCharDuration = 300 * time.Millisecond
```
SeBBBe, line 268: It's very slow typing

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+	}
+	if !hasResult {
+		logEntry.LogDebugf("search%d result: NONE", id)
+	}
+}
+
+var spaceRegex = regexp.MustCompile(`\s+`)
+
+func parseSearchTerms(s string) []string {
```
SeBBBe, line 178: We should get rid of this now that the search terms list has been changed to an array. No need to support splitting a string, right?

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+					{
+						GroupType: "DatasetType",
+						Offset:    0,
+						Count:     intPtr(-1),
+					},
+				},
+				GroupItemOptions: []*enigma.SearchGroupItemOptions{{
+					GroupItemType: "Field",
+					Offset:        0,
+					Count:         intPtr(5),
```
SeBBBe, line 248: Why 5? Perhaps add a const or comment

scenario/smartsearch.go
DnlLrssn: also in general maybe some comments to the chosen times.

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+		TypingDuration:          time.Duration(len(textChunk)) * typeOneCharDuration,
+	})
+	return chunks, nil
+
+}
+
+func (chunks searchTextChunks) simulate() <-chan string {
+	textChan := make(chan string, len(chunks))
+	go func() {
+		currentText := ""
```
DnlLrssn, line 308: add function to recover panic and log error on actions state in the go routine. (there's a helper function for recover)

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+		for _, chunk := range chunks {
+			currentText = currentText + chunk.Text
+			time.Sleep(chunk.TypingDuration)
```
DnlLrssn, line 311: use helper waitfor to also include context to break wait time on context canceled

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+		return
+	}
+
+	sessionState.LogEntry.LogDebugf("search text chunks: %+v", searchTextChunks)
+
+	reset()
+
+	cnt := 0
+	for searchText := range searchTextChunks.simulate() {
+		cnt++
```
DnlLrssn, line 343: add check for context canceled not to continue looping on execution stopped. Also with added possible error on actionstate as per above comment, a break check on actionState.Failed.

DnlLrssn (commented):
Still missing the possibility to select based on search results.

scenario/smartsearch.go
atluq: Alright, will add comments. This is based on typing speed 40 words per minute and average english word length 4.7.
60 * 1000 / (40 * 4.7) ~= 300 ms

But with the other breaks i agree it is a bit slow.

scenario/smartsearch.go
atluq: No, it is used to separate the search text into terms, right before doing the ws rpc:s.

scenario/smartsearch.go
atluq: It looks constant in the client. I have not observed any other value.

I add comments on this too.

scenario/smartsearch.go
```diff
@@ -0,0 +1,350 @@
...
+func parseSearchTerms(s string) []string {
+	searchTerms := []string{}
+	sb := &strings.Builder{}
+	quoted := false
+	escaped := false
+	addSearchTerm := func() {
+		defer sb.Reset()
+		searchTerm := sb.String()
+		searchTerm = strings.TrimSpace(searchTerm)
+		searchTerm = spaceRegex.ReplaceAllString(searchTerm, " ")
```
DnlLrssn, line 187: regex is usually heavy and slow, if we feel we can't solve it with string split or similar might want to have this done in unmarshal or similar not have every user have to do it for every invoke.

SeBBBe: Oh yeah, didn't even notice that. Would be better if we could move the entire `parseSearchTerms` into running with Unmarshal. Problem is, that interferes with the chunk pre-processing step since that involves randomization per-user. If you can think of a way to re-architect it such that we don't need to do regex or splitting in the Execute function, that would be a worthwhile improvement.

atluq: Splitting must be done in the execute function because of randomized typing as @SeBBBe said. The regex on the other hand is not that important, so I'll remove it or replace it with something more efficient. I noticed that if I wrote `"hello     world"` it became one search term `hello world`.

SeBBBe: Isn't it possible to run parseSearchTerms in the unmarshal and then run some lighter weight logic in the Execute? Since the chunk will always consist of x whole search terms, and possibly ending in a cut off one.

SeBBBe:
I still see potential for performance optimization by having all the string parsing in pre-processing but given the moderate rate of throughput, this should be good enough!

Are you working on implementing select?

atluq:
> I still see potential for performance optimization by having all the string parsing in pre-processing but given the moderate rate of throughput, this should be good enough!
> 
> Are you working on implementing select?

scenario/smartsearch.go
```diff
@@ -0,0 +1,462 @@
...
+	"github.com/qlik-oss/gopherciser/enummap"
+	"github.com/qlik-oss/gopherciser/helpers"
+	"github.com/qlik-oss/gopherciser/logger"
+	"github.com/qlik-oss/gopherciser/session"
+)
+
+const (
+	// typeOneCharDuration is based on:
+	//
+	// typing speed: 50 words per miniute
```
DnlLrssn, line 25: * minute

also general note (no need to change anything) "word" is extremely language specific and not a good metric, it's better to base it on Characters per minute as it would be language independent.

scenario/helpers.go
```diff
@@ -121,3 +123,22 @@ func (getVar varReq) WithCache(vc *enigmahandlers.VarCache) varReq {
...
+		return 0, errors.WithStack(err)
+	}
+	if delay < time.Nanosecond {
+		return 0, errors.New("timer delay not set")
+	}
+
+	select {
+	case <-ctx.Done():
+	case <-time.After(delay):
+	}
```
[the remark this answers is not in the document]
atluq: fixed

scenario/smartsearch.go
atluq: fixed

DnlLrssn (approved):

docs/settingup.md
```diff
@@ -1821,6 +1821,171 @@ Create and execute a `changesheet` action for each sheet in an app. This can be
...
+    "label": "one term search",
+    "settings": {
+        "searchtextlist": [
+            "term1"
+        ]
+    }
+}
+```
+
+#### Search with two search terms
```
SeBBBe, line 1866: Two search terms, but the label below is "one term search". I think the difference between terms separated by space and separated as an array needs to be clearly explained as I suspect it may be confusing.

scenario/smartsearch.go
```diff
@@ -0,0 +1,477 @@
...
+		return "NONE"
+	}
+	quoted := []string{}
+	for _, v := range strList {
+		quoted = append(quoted, quote(v))
+	}
+	return strings.Join(quoted, ",")
+}
+
+func ternary(cond bool, ifCondTrue string, ifCondFalse string) string {
```
SeBBBe, line 190: 😂

scenario/smartsearch.go
```diff
@@ -0,0 +1,477 @@
...
+			searchTermsMatched := []string{}
+			for _, stIndex := range item.SearchTermsMatched {
+				searchTermsMatched = append(searchTermsMatched, searchResult.SearchTerms[stIndex])
+			}
+			matches := []string{}
+			for _, match := range item.ItemMatches {
+				matches = append(matches, match.Text)
+			}
+			hasResult = true
+			logEntry.LogDebugf(`search%d result group[%d] matching term%s %s in %s %s: %s`, id, sgIdx,
```
SeBBBe, line 225: I like the exhaustive debug logging!

SeBBBe (approved):

scenario/smartsearch.go
atluq: ternary ? 😃 : rollYourOwn

scenario/smartsearch.go
```diff
@@ -0,0 +1,477 @@
...
+	searchTextChunks, err := newSearchTextChunks(rand, searchText, pasteSearchText)
+	if err != nil {
+		actionState.AddErrors(err)
+		return nil
+	}
+	reset()
+	defer sessionState.Wait(actionState)
+	cnt := 0
+	var searchResult **enigma.SearchResult
+	for searchText := range searchTextChunks.simulate(context.Background(), actionState.AddErrors) {
```
DnlLrssn, line 425: don't use background context

scenario/smartsearch.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+func (chunks searchTextChunks) simulate(ctx context.Context, onErrors func(err ...error)) <-chan string {
+	textChan := make(chan string, len(chunks))
+	go func() {
+		panicErr := helpers.RecoverWithErrorFunc(
+			func() {
+				defer close(textChan)
+				onContextDone := func() {
+					onErrors(errors.Wrap(ctx.Err(), "smart search typing simulation stopped"))
```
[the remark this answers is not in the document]
atluq: removed the reporting

scenario/smartsearch.go
atluq: what context then?

DnlLrssn: BaseContext from session would fit e.g. so session can be cancelled without the entire program is.

atluq: ahh, got a bit blind by my own code here, in my head it already was that one 🙈

atluq: should have used context.TODO() as a placeholder

DnlLrssn (approved):

PR #304, opened 2022-02-24
Bugfix of root cause for addTLS panic

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Bugfix of root cause for addTLS panic

**Info**

### DIFF
--- a/session/resthandler.go
+++ b/session/resthandler.go
@@ -125,16 +125,6 @@ var (
 	dnsResolver = &dnscache.Resolver{}
 )
 
-func init() {
-	go func() {
-		t := time.NewTicker(10 * time.Minute)
-		defer t.Stop()
-		for range t.C {
-			dnsResolver.Refresh(true)
-		}
-	}()
-}
-
 // NewRestHandler new instance of RestHandler
 func NewRestHandler(ctx context.Context, size int, trafficLogger enigma.TrafficLogger, headerjar *HeaderJar, virtualProxy string, timeout time.Duration) *RestHandler {
 	return &RestHandler{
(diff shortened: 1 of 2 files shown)


session/resthandler.go
```diff
@@ -227,12 +217,13 @@ func DefaultClient(allowUntrusted bool, state *State) (*http.Client, error) {
...
 						}
-						conn, err := dialer.DialContext(ctx, network, net.JoinHostPort(ip, port))
+						var conn net.Conn
+						conn, err = dialer.DialContext(ctx, network, net.JoinHostPort(ip, port))
 						if err == nil {
 							return conn, nil
 						}
 					}
-					return nil, nil
+					return nil, err
```
SeBBBe, line 226: This was the cause of the panic? Nil conn?

SeBBBe (approved):

session/resthandler.go
DnlLrssn: yupp

session/resthandler.go
```diff
@@ -125,16 +125,6 @@ var (
 	dnsResolver = &dnscache.Resolver{}
 )
 
-func init() {
-	go func() {
-		t := time.NewTicker(10 * time.Minute)
-		defer t.Stop()
-		for range t.C {
-			dnsResolver.Refresh(true)
```
SeBBBe, line 133: Why was this removed?

DnlLrssn: doesn't do much since it will be done on a cache fail anyways, and just increased systemcalls. Not directly related to this bugfix, but was removed during debugging and I saw no reason to re-add it.

PR #306, opened 2022-03-02
use <hr> instead of --- in markdown documentation for separator

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

makes output by pandoc look better + still works with github

**Info**


SeBBBe (approved):
Can only see remaining ---s in testdata, so looks good! Same changes need for qlik-trial/gopherciser, I assume

PR #309, opened 2022-03-10
Use default thinktime if default thinktime settings or nil

atluq:
This affects askhubadvisor and smartsearch. The reason for making the
check for deafults and nil is that the gui produces a json object with
static thinktime 0 by default, which is an invalid thinktime.

**Checklist**

- [x] tests added
- [x] documentation updated

### DIFF
--- a/scenario/helpers.go
+++ b/scenario/helpers.go
@@ -172,15 +172,10 @@ func (execute executeFunc) Validate() ([]string, error) {
 	return nil, nil
 }
 
-var internalSetThinkTimeError = errors.New("internal error: can not set thinkTime if **ThinkTimeSettings is nil")
-
-func setThinkTimeIfNotSet(thinkTime **ThinkTimeSettings, fallback *ThinkTimeSettings) error {
-	if thinkTime == nil {
-		return internalSetThinkTimeError
+func thinkTimeWithFallback(thinkTime *ThinkTimeSettings, fallback ThinkTimeSettings) *ThinkTimeSettings {
+	if thinkTime == nil || thinkTime.Type == helpers.StaticDistribution && thinkTime.Delay == 0 {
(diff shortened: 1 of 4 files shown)


DnlLrssn (approved):

SeBBBe (commented):
I like this solution! I guess the main drawback is it will no longer be possible to use an empty thinktime, if that's even a use case. It will also affect existing scripts that rely on the default value.

In that case, it might still be a good idea to have a "none" mode. Or possibly allow 0 values in thinktimesettings which resolves to a thinktime setting without any delay, but that's more up to Daniel to decide as it's his baby :)

SeBBBe:
On the other hand, the behavior contained within this PR might even be more desirable even disregarding the GUI. It requires less configuration by the script author while still providing a reasonable default thinktime.

scenario/helpers.go
```diff
@@ -172,15 +172,10 @@ func (execute executeFunc) Validate() ([]string, error) {
...
-var internalSetThinkTimeError = errors.New("internal error: can not set thinkTime if **ThinkTimeSettings is nil")
-
-func setThinkTimeIfNotSet(thinkTime **ThinkTimeSettings, fallback *ThinkTimeSettings) error {
-	if thinkTime == nil {
-		return internalSetThinkTimeError
-	}
-	if *thinkTime == nil || (*thinkTime).Type == helpers.StaticDistribution && (*thinkTime).Delay == 0 {
-		*thinkTime = fallback
-		return nil
+func thinkTimeWithFallback(thinkTime *ThinkTimeSettings, fallback ThinkTimeSettings) *ThinkTimeSettings {
```
SeBBBe, line 175: I like this better!

SeBBBe (approved):

PR #310, opened 2022-03-11
Fix internal thinktime defaults

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Fix internal thinktime GUI defaults

- Marshal default values if zero value in json
- Set smartsearch default to static delay 1s for less confusion
- Change action internal thinktimes to values, no null check needed because zero value is already checked.
- Add GUIDefaults interface

**Info**

### DIFF
--- a/docs/settingup.md
+++ b/docs/settingup.md
@@ -432,7 +432,7 @@ Perform a query in the Qlik Sense hub insight advisor.
 * `app`: Optional name of app to pick in followup queries. If not set, a random app is picked.
 * `saveimages`: Save images of charts to file.
 * `saveimagefile`: File name of saved images. Defaults to server side file name. Supports [Session Variables](https://github.com/qlik-trial/gopherciser-oss/blob/master/docs/settingup.md#session-variables).
-* `thinktime`: Settings for the `thinktime` action, which is automatically inserted before each followup.
+* `thinktime`: Settings for the `thinktime` action, which is automatically inserted before each followup. Defaults to a uniform distribution with mean=8 and deviation=4.
   * `type`: Type of think time
       * `static`: Static think time, defined by `delay`.
       * `uniform`: Random think time with uniform distribution, defined by `mean` and `dev`.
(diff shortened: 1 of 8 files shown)


scenario/smartsearch.go
```diff
@@ -167,6 +165,13 @@ func (settings *SmartSearchSettings) UnmarshalJSON(bytes []byte) error {
 	}
 	return nil
 }
+func (settings SmartSearchSettings) MarshalJSON() ([]byte, error) {
+	settings.SelectionThinkTime = thinkTimeWithFallback(
+		settings.SelectionThinkTime,
+		smartSearchDefaultThinktimeSettings,
+	)
+	return json.Marshal(settings)
+}
```
DnlLrssn, line 174: infinite loop

docs/settingup.md
```diff
@@ -432,7 +432,7 @@ Perform a query in the Qlik Sense hub insight advisor.
 * `app`: Optional name of app to pick in followup queries. If not set, a random app is picked.
 * `saveimages`: Save images of charts to file.
 * `saveimagefile`: File name of saved images. Defaults to server side file name. Supports [Session Variables](https://github.com/qlik-trial/gopherciser-oss/blob/master/docs/settingup.md#session-variables).
-* `thinktime`: Settings for the `thinktime` action, which is automatically inserted before each followup.
+* `thinktime`: Settings for the `thinktime` action, which is automatically inserted before each followup. Defaults to a uniform distribution with mean=8 and deviation=4.
```
SeBBBe, line 435: Wrong!

SeBBBe (changes requested):

scenario/smartsearch.go
atluq: 👍

docs/settingup.md
atluq: Note for hubadvisor. https://github.com/qlik-oss/gopherciser/blob/fb58db4096402ae78fb7f1980ff40f1828ebe99a/scenario/askhubadvisor.go#L60

I kept this to not change results of old scripts with implicit thinktime.

DnlLrssn (approved):

SeBBBe (approved):

PR #324, opened 2022-04-21
Autochart selections check

HvonP:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Autochart selections check

### DIFF
--- a/session/autocharthandler.go
+++ b/session/autocharthandler.go
@@ -40,11 +40,14 @@ func (instance *AutoChartInstance) SetObjectAndEvents(sessionState *State, actio
 }
 
 func (instance *AutoChartInstance) GetObjectDefinition(objectType string) (string, senseobjdef.SelectType, senseobjdef.DataDefType, error) {
+	selectType := senseobjdef.SelectTypeUnknown
+	selectPath := ""
 	// Checking if selectable
-	if instance.ObjectDef.Select == nil {
-		return "", senseobjdef.SelectTypeUnknown, senseobjdef.DataDefUnknown, errors.New("Autochart object has no selectable option")
+	if instance.ObjectDef.Select != nil {
+		selectPath = instance.ObjectDef.Select.Path
+		selectType = instance.ObjectDef.Select.Type
 	}
-	return instance.ObjectDef.Select.Path, instance.ObjectDef.Select.Type, instance.ObjectDef.DataDef.Type, nil
+	return selectPath, selectType, instance.ObjectDef.DataDef.Type, nil
 }
 
 func (instance *AutoChartInstance) handleAutoChart(sessionState *State, actionState *action.State, autochartObj *enigmahandlers.Object, autochartGen *enigma.GenericObject) {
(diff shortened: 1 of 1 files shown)


DnlLrssn (commented):
GetObjectDefinition is done for more than select I think?

session/autocharthandler.go
```diff
@@ -40,8 +40,12 @@ func (instance *AutoChartInstance) SetObjectAndEvents(sessionState *State, actio
...
 
 func (instance *AutoChartInstance) GetObjectDefinition(objectType string) (string, senseobjdef.SelectType, senseobjdef.DataDefType, error) {
-	// Todo error checking
-	return instance.ObjectDef.Select.Path, instance.ObjectDef.Select.Type, instance.ObjectDef.DataDef.Type, nil
+	// Checking if selectable
+	if instance.ObjectDef.Select != nil {
+		return instance.ObjectDef.Select.Path, instance.ObjectDef.Select.Type, instance.ObjectDef.DataDef.Type, nil
+	}
+	//Nothing selectable
+	return "", senseobjdef.SelectTypeUnknown, senseobjdef.DataDefUnknown, errors.New("Autochart object has no selectable option")
```
atluq, line 48: Is datadef always nil when select is?

I'd also flip the table and return the error in the if statement

DnlLrssn: no datadef should always be set. and as per previous comment, think this is the wrong place to do the check, changesheet etc would give an error getting the data because there's no "select"  I think?

SeBBBe (approved):

SeBBBe (approved):

DnlLrssn (changes requested):
will go ahead and request change since others approved. function should not return error just because object is not selectable

DnlLrssn:
see default handler for a solution:

> func (instance *DefaultHandlerInstance) GetObjectDefinition(objectType string) (string, senseobjdef.SelectType, senseobjdef.DataDefType, error) {
	def, defErr := senseobjdef.GetObjectDef(objectType)
	if defErr != nil {
		return "", senseobjdef.SelectTypeUnknown, senseobjdef.DataDefUnknown, errors.Wrapf(defErr, "Failed to get object<%s> selection definitions", objectType)
	}

	if validateErr := def.Validate(); validateErr != nil {
		return "", senseobjdef.SelectTypeUnknown, senseobjdef.DataDefUnknown, errors.Wrapf(validateErr, "Error validating object<%s> selection definitions<%+v>", objectType, def)
	}

	selectType := senseobjdef.SelectTypeUnknown
	selectPath := ""

	if def.Select != nil {
		selectPath = def.Select.Path
		selectType = def.Select.Type
	}

	return selectPath, selectType, def.DataDef.Type, nil
}

SeBBBe:
Right. It should just return `senseobjdef.SelectTypeUnknown` and no error in that case. Look at how it's used in randomaction.

DnlLrssn:
> Right. It should just return `senseobjdef.SelectTypeUnknown` and no error in that case. Look at how it's used in randomaction.

correct, and more important , return the correct datadef type

HvonP:
> > Right. It should just return `senseobjdef.SelectTypeUnknown` and no error in that case. Look at how it's used in randomaction.
> 
> correct, and more important , return the correct datadef type

Improved accordingly now

SeBBBe (approved):
This logic matches the default handler, looks good!

DnlLrssn (approved):

atluq (approved):
👍

PR #341, opened 2022-06-16
Allow different headers by method type

SeBBBe:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Allow adding different headers by HTTP method type
* Add origin header to all current method types except GET

**Info**

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -169,11 +169,15 @@ func (connectionSettings *ConnectionSettings) GetHeaders(state *session.State, e
 
 	state.HeaderJar.SetHeader(host, header)
 
+	restHost, err := connectionSettings.GetRestUrl()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 	dataMutationHeaders := state.HeaderJar.GetHeaderForMethod(host, session.POST)
 	if dataMutationHeaders == nil {
 		dataMutationHeaders = make(http.Header, len(connectionSettings.Headers))
 	}
-	dataMutationHeaders.Add("Origin", host)
+	dataMutationHeaders.Add("Origin", restHost)
 	state.HeaderJar.SetHeaderForMethods(host, dataMutationHeaders, []session.RestMethod{session.POST, session.DELETE, session.PUT, session.PATCH})
 
 	return header, nil
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -169,6 +169,13 @@ func (connectionSettings *ConnectionSettings) GetHeaders(state *session.State, e
 
 	state.HeaderJar.SetHeader(host, header)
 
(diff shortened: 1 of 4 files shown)


connection/connection.go
```diff
@@ -169,6 +169,17 @@ func (connectionSettings *ConnectionSettings) GetHeaders(state *session.State, e
 
 	state.HeaderJar.SetHeader(host, header)
 
+	restHost, err := connectionSettings.GetRestUrl()
```
DnlLrssn, line 172: can't remember, what's the difference between above `host` and the `GetRestUrl`, with and without protocol?

session/headerjar.go
```diff
@@ -1,30 +1,67 @@
...
+		headers:        &sync.Map{},
+		allRestMethods: allRestMethods,
 	}
 }
 
 // SetHeader set header for a particular host
 func (hj *HeaderJar) SetHeader(host string, header http.Header) {
-	hj.headers.Store(host, header)
+	if len(hj.allRestMethods) == 0 {
+		panic("HeaderJar may only be initialized using NewHeaderJar()")
```
DnlLrssn, line 43: never like seeing `panic` as it will affect all users, but guess the purpose if this is to catch programming errors early.

DnlLrssn (approved):

connection/connection.go
SeBBBe: Exactly. Tested first with host, resulted in errors so protocol is definitely required here.

atluq (approved):

scheduler/scheduler.go
```diff
@@ -278,15 +278,19 @@ func (sched *Scheduler) StartNewUser(ctx context.Context, timeout time.Duration,
 }
 
 func setupRESTHandler(sessionState *session.State, connectionSettings *connection.ConnectionSettings) error {
-	headers, err := connectionSettings.GetHeaders(sessionState, "")
+	host, err := connectionSettings.GetHost()
```
DnlLrssn, line 281: shouldn't even need the GetHost(), right?

SeBBBe: It's used on line 290

DnlLrssn: wait, ok so just `_, err = connectionSettings.GetHeaders(sessionState, "")`  wouldn't be enough?

DnlLrssn: as that one does GetHost function if host parameter is `""`

DnlLrssn: didn't see the double usage before, had accidentally switched to unified diff in github which makes things hard  to read.

SeBBBe: Ah yes I see that now.

DnlLrssn (approved):

PR #344, opened 2022-06-21
circleci: Test go 1.18 & remove 1.15

atluq:


DnlLrssn (approved):
update go.mod to require 1.16 too maybe?

PR #355, opened 2022-08-11
Add option to ignore publishing already published sheets

atluq:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Add option to ignore publishing already published sheets

- Add some sheet access debug logging
- Decorate publish sheet error

**Info**

Publishing a public sheet give a generic access denied error, which was hard to trace. Some exported apps can have public sheets when uploaded again, which lead to the finding of this issue. I have not seen this behavior until this week, have you guys seen it before or is it new?

### DIFF
--- a/scenario/publishsheet.go
+++ b/scenario/publishsheet.go
@@ -67,7 +67,10 @@ func (publishSheetSettings PublishSheetSettings) Execute(sessionState *session.S
 	publishAction := func(sheet *senseobjects.Sheet, ctx context.Context) error {
 		if publishSheetSettings.IgnorePublished {
 			if sheet.Layout.Meta.Published {
-				sessionState.LogDebugf(`not publishing sheet<%s> "%s" since it is already published`, sheet.ID, sheet.Layout.Meta.Title)
+				sessionState.LogDebugf(
+					`not publishing sheet<%s> "%s" since it is already published`,
+					sheet.ID, sheet.Layout.Meta.Title,
+				)
 				return nil
 			}
 		}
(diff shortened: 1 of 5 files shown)


SeBBBe:
> Some exported apps can have public sheets when uploaded again

Never seen that before!

scenario/publishsheet.go
```diff
@@ -64,7 +65,24 @@ func (publishSheetSettings PublishSheetSettings) Execute(sessionState *session.S
...
 
 	publishAction := func(sheet *senseobjects.Sheet, ctx context.Context) error {
-		return sheet.GenericObject.Publish(ctx)
+		if publishSheetSettings.IgnorePublished {
+			if sheet.Layout.Meta.Published {
+				sessionState.LogDebugf(`not publishing sheet<%s> "%s" since it is already published`, sheet.ID, sheet.Layout.Meta.Title)
+				return nil
+			}
+		}
+		sheetAccessLevel := "private"
```
SeBBBe, line 74: Ternary! 😭

scenario/publishsheet.go
```diff
@@ -64,7 +65,24 @@ func (publishSheetSettings PublishSheetSettings) Execute(sessionState *session.S
...
+		if sheet.Layout.Meta.Published {
+			sheetAccessLevel = "public"
+		}
+
+		sessionState.LogDebugf(`publishing %s sheet<%s> "%s"`, sheetAccessLevel, sheet.ID, sheet.Layout.Meta.Title)
+
+		return errors.Wrapf(
+			sheet.GenericObject.Publish(ctx),
```
SeBBBe, line 82: Kind of don't like this wrap-or-skip-if-nil, since it makes the success case a bit harder to read.

SeBBBe: maybe separate it into two lines,

err := sheet.GenericObject.Publish(ctx)

return errors.Wrapf(err, ...

SeBBBe (approved):
I don't really know how often this action is used as a sim vs prepper action. As a user I'd prefer having this be the default, but I can see both sides of the argument.

scenario/publishsheet.go
atluq: 😭 😭

scenario/publishsheet.go
atluq: Alrighty!

atluq:
> I don't really know how often this action is used as a sim vs prepper action. As a user I'd prefer having this be the default, but I can see both sides of the argument.

It would be a breaking change if it was the default, but to the better 🤷‍♂️

SeBBBe (approved):
LGTM

SeBBBe (approved):

DnlLrssn (approved):

PR #367, opened 2022-10-14
WS Reconnect changes

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add reconnect errors to current state if existing, instead of separate error log
* Only allow one reconnect function on websocket reconnect.
* Rewrite connect/re-connect to look at session_created, session_attached pushed messages

**Info**

### DIFF
--- a/wsdialer/wsdialer.go
+++ b/wsdialer/wsdialer.go
@@ -181,7 +181,18 @@ func (dialer *WsDialer) ReadMessage() (int, []byte, error) {
 		data = append(data, m.Payload...)
 	}
 
-	if err == io.EOF {
+	disconnected := false
+	switch err := err.(type) {
+	// We might want to check sub errors, but have seen both net.ErrClosed and os.SyscallError on disconnect to handling all OpErrors as disconnects for now
+	case *net.OpError:
+		disconnected = true
+	default:
+		if err == io.EOF {
+			disconnected = true
+		}
+	}
+
+	if disconnected {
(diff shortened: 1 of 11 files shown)


foh79 (approved):
Looks good to me but Anton should take a look

wsdialer/wsdialer.go
```diff
@@ -181,7 +181,18 @@ func (dialer *WsDialer) ReadMessage() (int, []byte, error) {
 		data = append(data, m.Payload...)
 	}
 
-	if err == io.EOF {
+	disconnected := false
+	switch err := err.(type) {
+	// We might want to check sub errors, but have seen both net.ErrClosed and os.SyscallError on disconnect to handling all OpErrors as disconnects for now
```
atluq, line 186: 👌

atluq (approved):
Great work!

Consider making a type `ConnectFunc` out of `func(bool) (string, error)` so it can be documented. I figured out what the bool and string was, but it would be great to have quick access to it when reading the code.

wsdialer/wsdialer.go
DnlLrssn: ah true would makes sense will do that.

atluq (approved):
Great docs on the connect func. It might be appropriate to use it in some more places.

foh79 (approved):

atluq (approved):
Looks good!

PR #381, opened 2022-12-29
Add in-mem Get app structure method

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add function to Get appstructure in mem.
* Categorize warnings from app structure with warning type returned when getting the structure.

**Info**

### DIFF
--- a/config/appstructure.go
+++ b/config/appstructure.go
@@ -27,9 +27,17 @@ import (
 )
 
 type (
+	AppStructureWarning struct {
+		ObjectID            string
+		ObjectType          string
+		ObjectVisualization string
+		Warningtype         string
+		Message             string
+	}
+
 	// AppStructureReport reports warnings and fetched objects for app structure
 	AppStructureReport struct {
-		warnings     []string
+		warnings     []AppStructureWarning
 		warningsLock sync.Mutex
 	}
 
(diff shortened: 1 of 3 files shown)


config/appstructure.go
```diff
@@ -40,10 +49,36 @@ type (
...
+	AppStructureWarningSnapshotError            = "SnapshotError"
+)
+
+// String implements Stringer interface
+func (warning AppStructureWarning) String() string {
+	return fmt.Sprintf("object<%s> type<%s> visualization<%s> warning: %s", warning.ObjectID, warning.ObjectType, warning.ObjectVisualization, warning.Message)
+}
+
+func (cfg *Config) getAppStructureScenario(includeRaw bool) ([]scenario.Action, map[string]*GeneratedAppStructure) {
+	appStructureMap := make(map[string]*GeneratedAppStructure, 1)
```
atluq, line 80: Why is this initialized with explicit capacity 1?

atluq (approved):
Looks good!

config/appstructure.go
DnlLrssn: just because it's the most common size which will be used. (unless running script structure command with a scenario containing multiple openapp actions). You think it'd be better to keep lazy grow?

atluq: Alright, sounds reasonable 👍

PR #388, opened 2023-02-23
API metrics optional

HvonP:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

API metrics optional with reworked metrics arguments

**Info**

Replaced metricsport with metricslevel and metricsaddress with metricstarget. This allows api metrics to be turned on or off at will

### DIFF
--- a/buildmetrics/buildmetrics.go
+++ b/buildmetrics/buildmetrics.go
@@ -1,3 +1,4 @@
+//go:build linux || darwin || windows
 // +build linux darwin windows
 
 package buildmetrics
@@ -11,12 +12,18 @@ import (
 )
 
 var (
-	enabled bool
+	enabled    bool
+	apiEnabled bool
 )
 
 // MetricEnabled returns whether metrics are enabled
 func metricEnabled() bool {
 	return enabled
+}
+
+// ApiMetricEnabled returns whether metrics are enabled for apis as well
+func apiMetricEnabled() bool {
+	return apiEnabled
 }
 
 func getLabel(action string, label string) string {
@@ -28,7 +35,7 @@ func getLabel(action string, label string) string {
 
 // ReportApiResult reports the duration for a specific API path and response code
 func ReportApiResult(action, label, path, method string, responseCode int, duration time.Duration) {
-	if metricEnabled() {
+	if metricEnabled() && apiMetricEnabled() {
 		actionlabel := getLabel(action, label)
 		resultString := strconv.Itoa(responseCode)
 		metrics.ApiCallDuration.WithLabelValues(actionlabel, path, method, resultString).Observe(duration.Seconds())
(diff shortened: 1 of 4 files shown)


cmd/execute.go
```diff
@@ -49,8 +49,8 @@ type (
 )
 
 var (
-	metricsPort      int
-	metricsAddress   string
```
DnlLrssn, line 53: should probably keep and deprecate these, if set, convert them to the new settings and print a warning they are deprectated. Also simplifies rolling out this breaking change.

buildmetrics/buildmetrics.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || darwin || windows
```
atluq, line 1: what side effect do this have?

atluq (commented):
Agree with previous comments

cmd/execute.go
```diff
@@ -305,21 +305,29 @@ func execute() error {
...
-			if err != nil {
-				return MetricError(fmt.Sprintf("failed to start prometheus : %s ", err))
-			}
-		} else {
-			//Prometheus metrics will be pushed to a pushgateway
-			err := buildmetrics.PushMetrics(ctx, metricsPort, metricsAddress, metricsLabel, metricsGroupings, scenario.RegisteredActions())
-			if err != nil {
-				return MetricError(fmt.Sprintf("failed to start prometheus : %s ", err))
-			}
+	switch metricsLevel {
```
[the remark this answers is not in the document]
HvonP: Yeah done

cmd/execute.go
HvonP: Good idea. Added a crude parser that automatically turns off API level metrics

cmd/execute.go
```diff
@@ -305,21 +351,62 @@ func execute() error {
...
-			//Prometheus metrics will be pulled from the endpoint /metrics
-			err := buildmetrics.PullMetrics(ctx, metricsPort, scenario.RegisteredActions())
-			if err != nil {
-				return MetricError(fmt.Sprintf("failed to start prometheus : %s ", err))
+	if metricsLevel != "" || metricsPort > 0 {
+		var metricsType MetricLevel
+		var err error
+
+		// Temporary conversion from legacy code metrics arguments to new
+		if metricsPort > 0 {
```
DnlLrssn, line 359: 👍 would be good to also print a warning to console for usage of deprecated parameter so the user knows.

cmd/execute.go
```diff
@@ -213,8 +257,10 @@ func init() {
 	AddLoggingParameters(executeCmd)
 
 	// Prometheus
-	executeCmd.Flags().IntVar(&metricsPort, "metrics", 0, "Export via http prometheus metrics.")
-	executeCmd.Flags().StringVar(&metricsAddress, "metricsaddress", "", "If set other than empty string then Push otherwise pull, will be appended by port.")
+	executeCmd.Flags().StringVar(&metricsLevel, "metricslevel", "", "Export via http prometheus metrics. 0-off, 1-Pull, 2-Push without api, 3 push with api")
```
DnlLrssn, line 260: update to reflect string names of enum, i.e. maybe something like
..."\n\t[0] nometrics (default)\n\t[1] pullmetric"
or similar so user can know something more understandable then the number can be written. 
it's ok write string like that, if you want to do it more advanced you can generate this help from the map like we do with loglevel

HvonP: Good suggestion, added

buildmetrics/buildmetrics.go
HvonP: It's that it only builds for those targets
For other targets there is a corresponding side implementation

Almost like an hardware/target interface

In this case there is a _JS file that nil implements JS bits

atluq: Cool, thanks!

PR #419, opened 2023-08-16
Apicompliance 1288 - Open App

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Changes for API compliance with latest Sense version

**Info**

### DIFF
--- a/session/sessionstate.go
+++ b/session/sessionstate.go
@@ -469,6 +469,17 @@ func (state *State) RegisterEvent(handle int,
 	})
 }
 
+// GetEvent for handle
+func (state *State) GetEventFunc(handle int) func(ctx context.Context, actionState *action.State) error {
+	state.eventMu.Lock()
+	defer state.eventMu.Unlock()
+	event, exist := state.events[handle]
+	if exist && event != nil {
+		return event.F
+	}
+	return nil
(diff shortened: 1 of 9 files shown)


SamiraNooran98 (approved):

session/sessionstate.go
```diff
@@ -469,6 +469,17 @@ func (state *State) RegisterEvent(handle int,
...
 }
 
+// GetEvent for handle
+func (state *State) GetEventFunc(handle int) func(ctx context.Context, actionState *action.State) error {
+	state.eventMu.Lock()
+	defer state.eventMu.Unlock()
+	event, exist := state.events[handle]
+	if exist && event != nil {
+		return event.F
+	}
```
atluq, line 479: minor: Could be simplified
```suggestion
	event := state.events[handle]
	if event != nil {
		return event.F
	}
```

atluq (approved):
Looks good!

session/sessionstate.go
DnlLrssn: having the `exists` helps if you ever need to debug, to distinguish between, exists with value null and doesn't exist.

PR #420, opened 2023-08-18
ignore error for GetScriptEx on openapp

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

ignore error for GetScriptEx request on openapp

**Info**


atluq (approved):
Right, we can not have that error each time. What happened will be in the traffic log. Might help with a debug log here too.

DnlLrssn:
sure debug log makes sense

PR #425, opened 2023-09-21
Expose artifact map in session variables

atluq:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Expose artifact map in session variables

This will enable using resource ids in templates, e-g:
```json
{
  "action": "setscriptvar",
  "settings": {
    "name": "MyAppId",
    "type": "string",
    "value": "{{.Artifacts.GetIDByTypeAndName \"app\" (print \"an-app-\" .Session)}}"
  }
}
```

**Info**

### DIFF
--- a/session/sessionstate.go
+++ b/session/sessionstate.go
@@ -132,6 +132,7 @@ type (
 		Thread     uint64
 		ScriptVars map[string]interface{}
 		Local      interface{}
+		Artifacts  *TemplateArtifactMap
 	}
 
 	ObjectHandlerInstance interface {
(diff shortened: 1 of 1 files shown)


DnlLrssn:
not directly related to PR. but is`print` actually needed in the golang template?

session/sessionstate.go
```diff
@@ -601,6 +602,20 @@ func (state *State) TriggerContextChanges(ctx context.Context, actionState *acti
...
 }
 
+type (
+	TemplateArtifactMap struct {
+		artifactMap *ArtifactMap
+	}
+)
+
+func (artifacts *TemplateArtifactMap) GetIDByTypeAndName(artifactType, name string) (string, error) {
+	artifact, err := artifacts.artifactMap.Lookup(artifactType, name, ArtifactEntryCompareTypeName)
```
DnlLrssn, line 612: probably should have a check if artifacts.artifactMap != nil , if it for some reason was nil when added in GetSessionVariable

DnlLrssn (commented):
Maybe should add a function to lookup name from ID too? 
Otherwise, yea think this should work and be a good addition.

atluq:
> not directly related to PR. but is`print` actually needed in the golang template?

There might be another way to concatenate strings inside of the braces. I wanted to include an example where `.Session` was used in the name since it is a common practice.

session/sessionstate.go
atluq: Yeah, I'll add that as a safeguard if artifactmap is not set for any reason.

atluq:
> Maybe should add a function to lookup name from ID too? Otherwise, yea think this should work and be a good addition.

Would be a simple addition. Do you have a use case example? Can't come up with any at the top of my head.

DnlLrssn:
> > Maybe should add a function to lookup name from ID too? Otherwise, yea think this should work and be a good addition.
> 
> Would be a simple addition. Do you have a use case example? Can't come up with any at the top of my head.

E.g. if you want to use the name in a POST request, e.g. a title.

DnlLrssn:
> > not directly related to PR. but is`print` actually needed in the golang template?
> 
> There might be another way to concatenate strings inside of the braces. I wanted to include an example where `.Session` was used in the name since it is a common practice.

Did a quick check seems it's necessary to do it something like that.

PR #440, opened 2023-12-07
Improvements to layout container handling

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add handling for show conditions in layout container
* Add handling for master objects in layout container

**Info**


helpers/fuzzybool.go
```diff
@@ -0,0 +1,66 @@
...
+
+type (
+	// FuzzyBool resolves boolean sent as strings, integers, float64 or boolean, json unmarshal defaults to true
+	FuzzyBool bool
+)
+
+// UnmarshalJSON FuzzyBool
+func (sb *FuzzyBool) UnmarshalJSON(arg []byte) error {
+	if sb == nil {
+		return nil
```
atluq, line 17: Is there some reason I am missing for not throwing an error here? Better fail fast.

helpers/fuzzybool.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	tmpMap := make(map[string]interface{})
+	tmpArray := []byte(`{"val":`)
+	tmpArray = append(tmpArray, arg...)
+	tmpArray = append(tmpArray, byte('}'))
```
atluq, line 23: Can't this be solved without this map trick?

With something like this:
``` golang
	var val interface{}
	err := json.Unmarshal(jsonRaw, &val)
	switch val.(type) {
	}
```

helpers/fuzzybool_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/qlik-oss/gopherciser/helpers"
+)
+
+func TestFuzzyBool_UnmarshalJSON(t *testing.T) {
+	tests := []struct {
```
atluq, line 11: Great!

helpers/fuzzybool_test.go
```diff
@@ -0,0 +1,96 @@
...
+			args:    []byte(fmt.Sprintf("%f", helpers.DefaultEpsilon)),
+			wantErr: false,
+			wantSb:  false,
+		},
+		{
+			name:    "Test float 1.0",
+			args:    []byte(`1.0`),
+			wantErr: false,
+			wantSb:  true,
+		},
```
atluq, line 82: Maybe add a null test

session/objecthandling.go
```diff
@@ -50,7 +50,12 @@ func init() {
 	if err := GlobalObjectHandler.RegisterHandler("container", &ContainerHandler{}, false); err != nil {
 		panic(err)
 	}
-
+	if err := GlobalObjectHandler.RegisterHandler("sn-layout-container", &LayoutContainerHandler{}, false); err != nil {
+		panic(err)
+	}
+	if err := GlobalObjectHandler.RegisterHandler("masterobject", &MasterObjectHandler{}, false); err != nil {
+		panic(err)
+	}
```
atluq, line 58: Old pattern, but just to double check: Is this safe as long as it  does not panic in the unit tests?

atluq (commented):
Looks good! A few questions and comments before approving...

helpers/fuzzybool.go
DnlLrssn: hmm that actually seems to work now. It used to be interface{} was not an allowed type to allocate for json.unmarshal, when did that change?

DnlLrssn: changed it, had missed that was allowed now. Also changed to use go-json instead.

session/objecthandling.go
DnlLrssn: not sure what you mean with "safe" it should only ever panic during start up of the program, not otherwise during runtime. And yes that would include unit tests for that package.

DnlLrssn: i.e. it's very "fail fast"

helpers/fuzzybool.go
DnlLrssn: is it an error? I see it as "nothing to unmarshal" ?

helpers/fuzzybool_test.go
DnlLrssn: good point, it didn't support "null" or "", so added that too!

helpers/fuzzybool.go
atluq: I think of it like there is no place to put the unmarshalled raw json.

helpers/fuzzybool_test.go
```diff
@@ -0,0 +1,114 @@
...
+			args:    []byte(``),
+			wantErr: false,
+			wantSb:  true,
+		},
+		{
+			name:    "Test null",
+			args:    []byte(`null`),
+			wantErr: false,
+			wantSb:  true,
+		},
```
atluq, line 28: Are these expected to be `true`? Seems counter intuitive.

session/objecthandling.go
atluq: Cool! 😎

helpers/fuzzybool.go
atluq: It is a minor, so if we disagree it can stay as is

helpers/fuzzybool_test.go
DnlLrssn: yes as per comment in begining of file, default is true. Mostly used for "conditions" etc where all values expect explicitly false is true.

DnlLrssn: i.e. null = no condition has been set and can thus be considered fulfilled

PR #454, opened 2024-02-09
Bugfix artifactmap using GUID from list or file

DnlLrssn:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

* Add Lookup in artifactmap also for guid from list and guid from file
* Add test build for Go 1.22

**Info**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,6 +186,9 @@ workflows:
       - test-go-version:
           name: "Test Go 1.21"
           version: "1.21"
+      - test-go-version:
+          name: "Test Go 1.22"
+          version: "1.22"
       - publish-github:
           requires:
             - build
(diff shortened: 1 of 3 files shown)


atluq (commented):
It makes sense to always return a real existing entry and not mix with fake entries, but I isn't that one point of the GUID app selection to not need to run an action to populate the artifact map before running e.g. `openapp`?

DnlLrssn:
> It makes sense to always return a real existing entry and not mix with fake entries, but I isn't that one point of the GUID app selection to not need to run an action to populate the artifact map before running e.g. `openapp`?

openapp would still work that way, other actions requiring data from the artifact map returns an error if it's not a "real" one when using GUID's. And yes the "fake" entry is to allow for some misuse of the tool, but we shouldn't force other actions to fail to support a misuse scenario.

atluq (approved):
I missed that there was an artificial mapEntry returned from LookupAppGUID. Question answered. Looks good!

PR #456, opened 2024-02-27
Refactor `reload`

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Fix bug in GetURL function
* Refactor reload action + add function which could be called from other actions.

**Info**

### DIFF
--- a/scenario/reload.go
+++ b/scenario/reload.go
@@ -9,6 +9,7 @@ import (
 	"github.com/qlik-oss/enigma-go/v4"
 	"github.com/qlik-oss/gopherciser/action"
 	"github.com/qlik-oss/gopherciser/connection"
+	"github.com/qlik-oss/gopherciser/enigmahandlers"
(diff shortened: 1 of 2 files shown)


scenario/reload.go
```diff
@@ -91,74 +92,110 @@ func (settings ReloadSettings) Execute(sessionState *session.State,
...
+				// Send one last progress message
+				if err := sessionState.SendRequest(actionState, getProgress); err != nil {
+					return errors.Wrap(err, "Error during reload")
 				}
+				if settings.SaveLog {
+					progressMessage = fmt.Sprintf("%s%s", progressMessage, progress.PersistentProgress)
+				}
+				return nil
+			case <-sessionState.BaseContext().Done():
+				return nil
```
atluq, line 156: Maybe return error here

atluq (approved):
Looks good! Just one observation on the context done handling.

PR #457, opened 2024-02-27
Add support extending http requests with custom headers

atluq:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

Add support extending http client with custom headers

**Info**


session/resthandler.go
```diff
@@ -546,10 +606,15 @@ func (handler *RestHandler) QueueRequestWithCallback(actionState *action.State,
 			return
 		}
 
-		if errRequest = handler.performRestCall(handler.ctx, request, handler.Client, handler.headers.GetHeader(host)); errRequest != nil {
+		req, errRequest := newStdRequest(handler.ctx, request, logEntry, handler.headers.GetHeader(host))
+		if errRequest != nil {
 			WarnOrError(actionState, logEntry, failOnError, errors.WithStack(errRequest))
```
DnlLrssn, line 611: should we maybe fail the request here if it's set to failOnError before even sending it?

atluq: Kept it like this to keep the behavior compatible with previous code, but doing so felt kind of wrong to begin with, so I'm changing it to fail here anyways.

PR #474, opened 2024-05-08
Enhancement/improved user scheduler

DnlLrssn:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

* Increase buffer size for logchan
* Use timer instead of time.After to decrease time to GC (can be GC'd after .Stop is called)
* Improved scalability of user session creator loop

**Info**


atluq (approved):
Looks good!

PR #485, opened 2024-06-20
Option to truncate appstructure

DnlLrssn:
**Checklist**

- [ ] tests added
- [ ] documentation updated

**Squash commit message**

* Add option to truncate titles and descriptions in appstructure to a defined value to avoid unnecessary memory impact.
* Workaround for engine ws bug

**Info**

### DIFF
--- a/config/appstructure.go
+++ b/config/appstructure.go
@@ -48,6 +48,7 @@ type (
 		logEntry      *logger.LogEntry
 		report        AppStructureReport
 		structureLock sync.Mutex
+		truncateTo    int
 	}
 
 	// Appstructure and warnings in consumable form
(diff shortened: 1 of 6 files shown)


config/appstructure.go
```diff
@@ -987,3 +982,11 @@ func (report *AppStructureReport) AddWarning(warning AppStructureWarning) {
 
 	report.warnings = append(report.warnings, warning)
 }
+
+func truncateMeta(meta *appstructure.MetaDef, truncateTo int) {
+	if truncateTo < 1 {
```
HvonP, line 987: Fair, but isn't there a way to impose rules on input argument in config and set to a default if less than 1?

This is added security of course

HvonP (approved):
Looks good

config/appstructure.go
DnlLrssn: The default is to not truncate

PR #488, opened 2024-08-02
Enhancement add modulo func to sync template

pawel-fobka:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

Added modulo func to sync template

**Info**


generatedocs/data/extra/sessionvariables/description.md
```diff
@@ -96,4 +97,33 @@ The following functions are supported:
...
+  "settings": {
+    "subscriptiontype": "template-sharing",
+    "limitperpage": 100,
+    "appname": "PS-18566_Test_Levels_Pages- {{ modulo .Session 4}}",
+    "subscriptionmode": "random",
+  }
+}
+```
+
+Very similar case as above but apps have number suffix from 1 to 4. This can be habdled combining `modulo` and `add` functions. The cycle through the suffix number will be done in following order: 2, 3, 4, 1.
```
DnlLrssn, line 115:
```suggestion
Very similar case as above but apps have number suffix from 1 to 4. This can be handled combining `modulo` and `add` functions. The cycle through the suffix number will be done in following order: 2, 3, 4, 1.
```

DnlLrssn (approved):

PR #497, opened 2024-08-27
Bugfix: pending waiter

DnlLrssn:
**Checklist**

- [x] tests added
- [ ] documentation updated

**Squash commit message**

Fixes bug where something added to state.pending after emptying, but before emptying of resthandler.pending would not wait as expected.

**Info**


pending/pendinghandler.go
```diff
@@ -20,7 +20,7 @@ type (
 )
 
 // NewHandler new instance of PendingHandler
-func NewHandler(size int) Handler {
+func NewHandler() Handler {
```
atluq, line 23: `size`does not matter 😉

session/sessionstate_test.go
```diff
@@ -403,3 +404,42 @@ func TestStateTemplateArtifactMap(t *testing.T) {
 	})
 
 }
+
+func TestPendingWaiter(t *testing.T) {
+	counters := &statistics.ExecutionCounters{}
+	state := New(context.Background(), "", 120, nil, 1, 1, "", false, counters)
```
atluq, line 410: Did this simple test break the old one? If so, good! If not it would be a good idea to create such a test.

atluq (approved):
Looks good! One question and one joke ⬇️

PR #535, opened 2025-03-05
More checks on documentation generation

DnlLrssn:
**Checklist**

- [ ] tests added
- [x] documentation updated

**Squash commit message**

Additional checks to make `Verify generated` check fail when documentation is lacking.

* Change missing documentation to errors instead of warnings when generated.
* Add errors when action fields are missing `doc-key`.
* Add warning when action fields are missing `displayname`.
* Add missing documentation for parameter `changestream.stream`.

**Info**
image


pawel-fobka (approved):

generatedocs/pkg/doccompiler/checks.go
```diff
@@ -61,22 +58,23 @@ func checkAllConfigFieldsDocumented(data *docData) []error {
...
-			findings = append(findings, errors.Errorf(`config field "%s" has no description`, field))
+			findings = errors.Join(findings, fmt.Errorf(`config field "%s" has no description`, field))
 		}
 		if docEntry.Examples == "" {
-			findings = append(findings, errors.Errorf(`config field "%s" has no examples`, field))
+			findings = errors.Join(findings, fmt.Errorf(`config field "%s" has no examples`, field))
 		}
 	}
+
 	return findings
```
atluq, line 73: A single `errors.Join` at the end would create a flat structure that is easier to work with calling `Unwrap` on it.

generatedocs/pkg/doccompiler/checks.go
```diff
@@ -94,11 +92,137 @@ func checkAllActionsInGroup(data *docData) []error {
 		lenGroups := len(groups)
 		switch {
 		case lenGroups == 0:
-			findings = append(findings, errors.Errorf(`action "%s" does not belong to a group`, action))
+			findings = errors.Join(findings, fmt.Errorf(`action "%s" does not belong to a group`, action))
 		case lenGroups > 1:
-			findings = append(findings, errors.Errorf(`action "%s" belong to %d groups %v`, action, lenGroups, groups))
+			findings = errors.Join(findings, fmt.Errorf(`action "%s" belong to %d groups %v`, action, lenGroups, groups))
```
atluq, line 97: Same as obove

atluq (approved):
Looks good. Seems like best practice to keep the joined errors flat.

generatedocs/pkg/doccompiler/checks.go
DnlLrssn: ok, not sure what you mean, how do I add them to the findings without joining?

DnlLrssn: Unwrap always return `nil` on joined errors btw so not good idea to use. Commited a version with slightly reduced join nesting at least. Result print still same independent of join nesting.

PR #553, opened 2025-04-08
Autodetect privatekey type and update default signing algorithm during JSON unmarshal

DnlLrssn:
**Checklist**

- [x] tests added
- [x] documentation updated

**Squash commit message**

* When jwt connection is used, read private key and try to detect key type of `alg` has not been defined. 

**Info**


config/config_test.go
```diff
@@ -61,7 +39,7 @@ func TestConfig(t *testing.T) {
 			"virtualproxy": "myvp",
 			"security": true,
 			"jwtsettings": {
-				"keypath" : ` + string(tmpKeyFilePath) + `,
```
atluq, line 64: That looked weird 😵

atluq (approved):
Looks good!

config/config_test.go
DnlLrssn: used to be []byte not string

